PR #4, opened 2018-01-16
Add role for setup-sources

pcreech:
Add role for setup-sources

Fixes GH-3

### DIFF
--- a/obal/data/roles/setup_sources/tasks/annex.yml
+++ b/obal/data/roles/setup_sources/tasks/annex.yml
@@ -1,0 +1,32 @@
+---
+- set_fact:
+    source: "{{ item }}"
+    sourcebase: "{{ item | basename }}"
+  tags:
+    - always
+
+- set_fact:
+    sourcepath: "{{ package_dir  }}/{{ sourcebase }}"
+  tags:
+    - always
+
+- name: 'Source exists'
+  stat:
+    path: "{{ sourcepath }}"
+  register: source_exists
+
+- name: 'Search annex for web link'
+  shell: 'git annex whereis {{ sourcepath }} 2>/dev/null | grep -q " web:"'
+  args:
+    chdir: "{{ package_dir }}"
+  changed_when: False
+  register: search_annex
+  ignore_errors: yes
+  when: (not source_exists.stat.exists) or ( source_exists.stat.exists and source_exists.stat.islnk is defined and source_exists.stat.islnk )
+
+- name: 'Add url to annex for source'
+  shell: "git annex addurl {{ setup_sources_annex_options }} --file '{{ sourcebase }}' '{{ source }}'"
+  args:
+    chdir: "{{ package_dir }}"
+  ignore_errors: yes
+  when: search_annex|failed
--- a/obal/data/release_package.yml
+++ b/obal/data/release_package.yml
@@ -25,11 +25,10 @@
   roles:
     - git-annex-setup
     - ensure-package
+    - setup_sources
     - diff-brew-package
   tasks:
     - include: tasks/check_builds.yml
-
-    - include: tasks/setup_sources.yml
 
     - set_fact:
         tito_releasers: "{{ joined_releasers }}"
--- a/obal/data/roles/setup_sources/defaults/main.yaml
+++ b/obal/data/roles/setup_sources/defaults/main.yaml
@@ -1,0 +1,2 @@
+---
+setup_sources_annex_options: ""
--- a/obal/data/roles/setup_sources/tasks/main.yml
+++ b/obal/data/roles/setup_sources/tasks/main.yml
@@ -1,0 +1,28 @@
+---
+- set_fact:
+    package_dir: "{{ inventory_dir }}/packages/{{ inventory_hostname }}"
+  tags:
+    - always
+
+- name: get spec file
+  find:
+    pattern: "*.spec"
+    path: "{{ package_dir }}"
+  register: spec_file
+  tags:
+    - always
+
+- set_fact:
+    spec_file_path: "{{ spec_file.files[0].path }}"
+  tags:
+    - always
+
+- name: 'Extract Sources from spec'
+  shell: "spectool --list-files --sources {{ spec_file_path }} | awk '{print $2}'"
+  register: sources
+  changed_when: False
+  tags:
+    - always
+
+- include: annex.yml
+  with_items: "{{ sources.stdout_lines | list }}"
--- a/obal/data/scratch_build.yml
+++ b/obal/data/scratch_build.yml
@@ -5,10 +5,9 @@
   gather_facts: no
   roles:
     - git-annex-setup
+    - setup_sources
     - ensure-package
   tasks:
-    - include: tasks/setup_sources.yml
-
     - set_fact:
         tito_releasers: "{{ releasers | join(' ') | replace('dist-git', 'scratch') }}"
       tags:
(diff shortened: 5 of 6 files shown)


obal/data/roles/setup-sources/tasks/defaults/main.yaml
```diff
@@ -0,0 +1,3 @@
+---
+force: false
+annex_opts: ""
```
ehelms: Convention says to prefix variables with the role to namespace them and prevent clashes.

obal/data/roles/setup-sources/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
+---
+- set_fact:
+    pkg_dir: "{{ inventory_dir }}/packages/{{ inventory_hostname }}"
+    annex_opts: annex_opts
```
ehelms: I can live with the shortening of words if others are, but generally I tend towards explicit verbosity (e.g. package over pkg and options over opts).

ehelms: Does this need a tag always?

obal/data/roles/setup-sources/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+  register: spec_file
+  tags:
+    - always
+
+- set_fact:
+    spec_file_path: "{{ spec_file.files[0].path }}"
+  tags:
+    - always
+
+- name: 'Extract Source0 from spec'
```
ehelms: What happens if I have more than one source that is annexed? Rare, but possible.

obal/data/roles/setup-sources/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+    source0: "{{ source0.stdout }}"
+  tags:
+    - always
+
+- name: 'Extract sourcebase from Source0'
+  shell: "basename {{ source0 }}"
+  register: sourcebase
+  changed_when: False
+  tags:
+    - always
```
ehelms: I think you can combine this with the `set_fact` above using the `basename` filter Ansible provides

obal/data/roles/setup-sources/tasks/main.yml
pcreech: In this case, it will be ignored.  Based on my reading of the setup_sources.sh file I was basing this off of, this was the only case.  I can go back and add in support for multiple sources

ekohl: https://github.com/theforeman/foreman-packaging/blob/rpm/develop/setup_sources.sh does loop over the files

obal/data/roles/setup-sources/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
+---
+- set_fact:
+    package_dir: "{{ inventory_dir }}/packages/{{ inventory_hostname }}"
+    setup_sources_annex_options: setup_sources_annex_options
```
ekohl: Why is this needed?

obal/data/roles/setup-sources/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+  register: search_annex
+  ignore_errors: yes
+  when: setup_sources_force or ( source_exists.stat.islink is defined and source_exists.stat.islnk )
+
+- name: 'Add url to annex for source'
+  shell: 'git annex addurl {{ setup_sources_annex_options }} --file "{{ sourcebase }}" "{{ source0 }}"'
+  args:
+    chdir: "{{ package_dir }}"
+  ignore_errors: yes
+  when: search_annex.rc is defined and search_annex.rc != 0
```
ekohl: I'm still quite new to ansible but would ```search_annex|failed``` work?

obal/data/roles/setup-sources/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+  tags:
+    - always
+
+- name: 'Source exists'
+  stat: 
+    path: "{{ sourcepath }}"
+  register: source_exists
+
+- name: 'Search annex for web link'
+  shell: 'git annex whereis {{ sourcepath }} 2>/dev/null | grep -q " web"'
```
ekohl: foreman-packaging is a bit more specific in that it tests for ``` web:```

ekohl:
Should ```tasks/setup_sources.yml``` be removed too?

obal/data/roles/setup-sources/tasks/main.yml
evgeni: it's not, esp it sets the `setup_sources_annex_options` fact to the literal `setup_sources_annex_options` string

obal/data/roles/setup-sources/tasks/main.yml
evgeni: "our" version was ignoring other sources so far. how does the foreman code cope with non-remote files, or is `[ -h ... ]` enough to catch those?

evgeni:
When I added `-f` (aka `setup_sources_force`) to *our* copy of `setup_sources.sh`, the idea was that I want the script to do `git annex addurl` also for *new* files (so `[ -h ...]` was false).

now, a year later, I think the check also could have been:
```bash
  [ -h $d/$sourcebase ] || [ ! -e $d/$sourcebase ] || continue
```

is it a link? cool! does it not exist? cool. otherwise, don't touch this.

`git annex whereis` fails on a non-existing path, though, so we'd have to catch that.

obal/data/roles/setup-sources/tasks/main.yml
ekohl: I haven't seen issues with the actually included tarballs we have with bundled installations. According to ```man test```:
> -h FILE
> FILE exists and is a symbolic link (same as -L)

That implies it's never an issue with actual files.

obal/data/roles/setup-sources/tasks/main.yml
pcreech: Yeah, I think this snuck in there b/c I had the defaults in the wrong place and I didn't remove it

obal/data/roles/setup-sources/tasks/main.yml
ehelms: Yup.

ehelms:
General question given this is the first new role being contributed. I believe convention is to do roles with underscores in their names and thus their folder structure. Should we switch to that?

evgeni:
:+1: for python naming

pcreech:
Since it was simple to convert the spectool output into iterable items, I was able to refactor the role to support multiple source lines.

ehelms:
I am a bit confused on why the role moved under `obal/data/roles` ?

pcreech:
Re: move under obal/data/roles... looks like it was done as part of https://github.com/evgeni/obal/commit/b8469f51ab6da2dec99cead83a3d46d539ec1cf3

evgeni:
Yeah. That was the easiest way to distribute the roles as part of thr python  package. We can rethink that if you want to.

evgeni:
Yeah. That was the easiest way to distribute the roles as part of thr python  package. We can rethink that if you want to.

evgeni:
Yeah. That was the easiest way to distribute the roles as part of thr python  package. We can rethink that if you want to.

ehelms:
@evgeni I think it just surprised me as a change, and my first impression was how it hid them under a deep path that was not that discoverable or obvious - still noodling on it.

@ekohl Final review falls to you

ekohl:
> @evgeni I think it just surprised me as a change, and my first impression was how it hid them under a deep path that was not that discoverable or obvious - still noodling on it.

We could do the same as in foreman-infra and add a symlink at the root level.

obal/data/roles/setup_sources/tasks/annex.yml
```diff
@@ -0,0 +1,32 @@
+---
+- set_fact:
+    source0: "{{ item }}"
```
ekohl: Just source might be cleaner?

PR #9, opened 2018-01-18
Add copr scratch build

ehelms:
Note this requires @pcreech setup_sources role. I have only really tested copr scratch builds up to this point.


ehelms:
This switches tfm-ror51 to use it -- https://github.com/theforeman/tfm-ror51-packaging/pull/42/files

evgeni:
needs a rebase :)

evgeni:
this moves scratch builds to an own role, and thus kinda abandoning the release playbook, which uses the same `tito_release.yml` and `brew_wait_and_download.yml` sub-plays.

can we somehow make this a `build` role, that can do both, release and scratch builds?

ehelms:
Most likely, I will look into it. Scratch builds with Copr are at lot less trivial than with Koji unfortunately.

ehelms:
This has been fully updated to the new roles and integrating copr into it. Copr has an odd workflow for scratch builds that poses a challenge. Any  simplification design ideas would be appreciated.

pcreech:
@ehelms Looks like the tests blew up

obal/data/roles/build_package/tasks/copr.yml
```diff
@@ -0,0 +1,65 @@
...
+    - name: 'Build SRPM'
+      command: 'tito build --srpm --scl={{ scl }}'
+      args:
+        chdir: "{{ inventory_dir }}/packages/{{ inventory_hostname }}"
+      register: srpm_build
+      tags:
+        - always
+
+    - name: 'Run build'
+      command: "copr-cli build {{ copr_repo_name }} {{ srpm_build.stdout.split('Wrote: ')[1] }}"
```
ehelms: @evgeni I need a little testing help. This line here expects the previous task to have created an SRPM and stored that information in a variable that this extracts. The test framework with mockbin currently does not have any mocked output which results in `srpm_build.stdout` being empty string and then this fails during testing.

evgeni: Well, then let's mock `tito` a bit more, shall we?

```diff
diff --git a/tests/fixtures/mockbin/mockbin b/tests/fixtures/mockbin/mockbin
index a7223ac..22708ca 100755
--- a/tests/fixtures/mockbin/mockbin
+++ b/tests/fixtures/mockbin/mockbin
@@ -1,10 +1,29 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
+
+from __future__ import print_function
 
 import os
 import sys
+import argparse
 
 mockbin_log = os.environ.get('MOCKBIN_LOG', None)
 
 if mockbin_log:
     with open(mockbin_log, 'a') as logfile:
         logfile.write("%s\n" % sys.argv)
+
+prog = os.path.basename(sys.argv[0])
+
+if prog == 'tito':
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers()
+    parser_release = subparsers.add_parser('release')
+    parser_release.add_argument('dest')
+    parser_release.add_argument('--scratch', action='store_true')
+    parser_release.add_argument('-y', action='store_true')
+    parser_build = subparsers.add_parser('build')
+    parser_build.add_argument('--srpm', action='store_const', dest='output', const='Wrote: hello.src.rpm')
+    parser_build.add_argument('--scl', action='store')
+    args = parser.parse_args()
+    if 'output' in args and args.output:
+        print(args.output)
diff --git a/tests/test_functional.py b/tests/test_functional.py
index 177e86d..19a9667 100644
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -208,7 +208,7 @@ def test_obal_scratch_copr_hello():
         "['{bin}/copr-cli', 'create', 'copr-repo-scratch', '--chroot', 'epel-7-x86_64', '--description', 'Scratch Builds', '--unlisted-on-hp', 'on', '--repo', 'http://mirror.centos.org/centos/7/sclo/x86_64/rh/']",  # noqa: E501
         "['{bin}/copr-cli', 'edit-chroot', 'copr-repo-scratch/epel-7-x86_64', '--packages', 'scl-utils-build rh-ruby24-build']",  # noqa: E501
         "['{bin}/tito', 'build', '--srpm', '--scl=copr-scl']",
-        "['{bin}/copr-cli', 'build', 'copr-repo-scratch']"
+        "['{bin}/copr-cli', 'build', 'copr-repo-scratch', 'hello.src.rpm']"
     ]
     assert_mockbin_log(expected_log)
 
```

ehelms:
@evgeni Green with your patch! Thanks, please to review

evgeni:
I think this looks good.

The diff is a bit noisy with all that `s/'/"/`, but I don't mind too much.

Would love a second pair of eyes (@ekohl, @pcreech?) who understands copr better than I do.

obal/data/roles/copr_repo/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: 'Create copr scratch build repo'
+  command: >
+    copr-cli
+      create
```
pcreech: will this command fail out if the copr repo is already created? I.E. when you are building a release to an existing copr repo?

pcreech: Ah, nevermind.  That 'when: build_package_scratch|default(false)' at the end of the whole block prevents it from running in the non-scratch case

ehelms:
Here goes..

PR #12, opened 2018-01-18
add support for ansible tags

evgeni:


obal/__init__.py
```diff
@@ -77,6 +85,40 @@ def main():
...
+
+
+def generate_ansible_args(inventory_path, playbook_path, args):
+    limit = ':'.join(args.package)
+    ansible_args = [playbook_path, '--inventory', inventory_path, '--limit',
+                    limit]
+    for extra_var in args.extra_vars:
+        ansible_args.extend(["-e", extra_var])
+    if args.verbose:
+        ansible_args.append("-%s" % str("v" * args.verbose))
```
ekohl: Can also be written as ```"-" + "v" * args.verbose```

evgeni: yeah, I kinda think both are ugly :(

PR #15, opened 2018-01-18
properly exit on error

evgeni:
Fixes: GH-14

### DIFF
--- a/obal/__init__.py
+++ b/obal/__init__.py
@@ -5,6 +5,7 @@ from __future__ import print_function
 
 import argparse
 import os
+import sys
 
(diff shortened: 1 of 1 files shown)


ekohl:
I believe what ```sys.exit``` does is raise a ```SystemExit``` which you can actually test for if you want :)

evgeni:
yeah, but that would require having a playbook that fails intentionally during tests :)

ekohl:
I'd guess a playbook that doesn't exist would also cause an error or it could call ```/bin/false``` in a command. Pretty sure that should fail.

Issue #19, opened 2018-01-18
`-v` does not work as desired


evgeni:
seems not passed properly to ansible

PR #22, opened 2018-01-19
create an ansible Display so that the CLI can set verbosity

evgeni:
Fixes: GH-19

### DIFF
--- a/obal/__init__.py
+++ b/obal/__init__.py
@@ -7,6 +7,7 @@ import argparse
 import os
 import sys
 
+from ansible.utils.display import Display
 from ansible.inventory.manager import InventoryManager
 from ansible.parsing.dataloader import DataLoader
 from pkg_resources import resource_filename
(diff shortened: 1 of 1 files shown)


obal/__init__.py
```diff
@@ -152,4 +153,8 @@ def main():
 
 
 if __name__ == '__main__':
+    # this needs to be outside of main, as otherwise PlaybookCLI fails
+    # to set the verbosity correctly
+    display = Display()
```
ekohl: This variable is unused. Either you can remove the variable (not the actual function) or you should use it.

evgeni: It's a tad more complicated than this. `PlaybookCLI` expects an `display` object to exist at the global scope, and modifies its `verbosity` attribute. If there is none, it will create it, but then the settings are not propagated and you end up with non-verbose output.

ekohl: I meant that ```Display()``` should be sufficient if you don't use the ```display``` variable but if it needs to exist at the global scope that's a good comment to add. pylint would complain about this. An alternative is to use ```global display = Display()``` in ```main()``` since that's explicit.

evgeni:
```
    global display = Display()
                   ^
SyntaxError: invalid syntax
```

evgeni: ah, it works in two lines.

PR #26, opened 2018-01-23
add more tests for the args

evgeni:


tests/test_obal.py
```diff
@@ -38,9 +38,48 @@ def test_generate_ansible_args_testpackage():
     assert 'testpackage' in ansible_args
 
 
+def test_generate_ansible_args_verbose():
```
ekohl: Personally I like ```@pytest.mark.parametrize```:
```python
@pytest.mark.parametrize('cliargs,expected', [
  (['--verbose', 'scratch', 'testpackage'], ['testpackage', '-v']),
  (['-vvvv', 'scratch', 'testpackage'], ['testpackage', '-vvvv']),
])
def test_generate_ansible_args_vvvv(cliargs, expected):
    ansible_args = _test_generate_ansible_args(cliargs)
    for exp in expected:
        assert exp in ansible_args
```

I wonder if we should test for the full result to keep it simple. IMHO it'd be a better test.

evgeni: what do you mean for full result? `assert ansible_args == ['ansible-playbook', ...]`?

ekohl: Yes

evgeni: right. updated

PR #27, opened 2018-01-23
make coverage understand the tox paths

evgeni:
this ensures that the tox paths are merged to the real path and coveralls and co can report the coverage nicely (with highlighting etc)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[paths]
+source = obal/
+  .tox/*/lib/*/site-packages/obal/
+
+[run]
+source = obal/
+parallel = true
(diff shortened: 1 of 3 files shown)


ehelms:
Is it expected that coverage dropped?

evgeni:
no, I have no idea why it thinks it dropped

evgeni:
@ehelms coverage drop fixed.

.coveragerc
```diff
@@ -0,0 +1,7 @@
+[paths]
+source = .
+  .tox/*/lib/*/site-packages/
+
+[run]
+source = .
```
ekohl: Should source be ```obal``` so we don't count our tests as coverage?

evgeni: given we call pytest with `--cov={envsitepackagesdir}/obal`, only `obal` is really scanned. but I can try adopting it here too.

evgeni: yeah, works.

PR #32, opened 2018-01-24
import *all* ansible parts *after* setting ANSIBLE_CONFIG

evgeni:
otherwise the config is not honoured at all

Fixes: #29

### DIFF
--- a/obal/__init__.py
+++ b/obal/__init__.py
@@ -7,9 +7,6 @@ import argparse
 import os
 import sys
 
-from ansible.utils.display import Display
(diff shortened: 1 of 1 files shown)


obal/__init__.py
```diff
@@ -27,6 +24,8 @@
 
 
 def find_packages(inventory_path):
+    from ansible.inventory.manager import InventoryManager
```
ekohl: You could even move these inside the if statement.

evgeni: true

PR #33, opened 2018-01-24
Refactor building/releasing into a role

pcreech:
Refactor buildling/releasing into a role


obal/data/roles/build_package/tasks/brew_wait_and_download.yml
```diff
@@ -1,13 +1,10 @@
 ---
```
ehelms: I think we could rename this file `wait_and_download` ? Or split the waiting and the downloading up

ehelms:
Best practices for Ansible roles are to scope variables in a role by the role name. Might as well do that here since this is re-factoring it to a role. I realize this might break projects using this but better to rip the band-aid early.

obal/data/roles/build_package/tasks/brew_wait_and_download.yml
pcreech: Separate wait and download sounds good.  I wasn't sure how far to take it in this one since my next pr is to make it brew/koji agnostic

pcreech:
Will scope the variables.  The only one that really matters from elsewhere I believe is the tito_releasers, and that could be modified in another way.  Since more radical changes are being introduced, I might go ahead and add the brew/koji work here as well and get it done.

obal/data/roles/build_package/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- include_tasks: koji.yml
```
evgeni: `include_tasks: "{{ build_package_build_system }}.yml"`? that allows to drop the `when`

also please add a `name`

obal/data/roles/build_package/defaults/main.yaml
```diff
@@ -0,0 +1,4 @@
+---
+build_package_build_system: koji
+build_package_koji_command: 'koji'
```
evgeni: can this be set automatically depending on the value of `build_package_build_system`? Or maybe we can just use `build_package_build_system` directly? I guess it will be `koji`/`brew` where you want this distinction, and for all the other build systems, we'd probably just never load the `koji` related task files?

obal/data/roles/build_package/tasks/main.yml
pcreech: Yup

obal/data/roles/build_package/defaults/main.yaml
pcreech: build_package_build_system is to denote that we are using a koji-based workflow.

build_package_koji_command is to specify the exact koji command to use when interacting with the koji system (koji, brew, kkoji*, etc...).  This is a variable specific to using 'koji.yml'

* for a while, to denote between fedora's koji and katello's koji, I had a kkoji command on my system, which is supported by koji

evgeni: Right. I forgot about the kkoji stuff.

obal/data/release_package.yml
```diff
@@ -22,20 +22,17 @@
...
-    - setup_sources
-    - diff_brew_package
-  tasks:
-    - include: tasks/check_builds.yml
-
+  pre_tasks:
     - name: 'Set tito_releasers'
       set_fact:
-        tito_releasers: "{{ joined_releasers }}"
+        build_package_tito_releasers: "{{ releasers | join(' ')  }}"
```
ehelms: I think we could move this into the role

obal/data/roles/build_package/tasks/wait.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: 'Get {{ build_package_koji_command }} task ids'
+  set_fact:
+    build_package_koji_tasks: "{{ build_package_tito_release.stdout_lines \
+      | select('match', '^Created task:.*') \
+      | map('regex_replace', '^Created task:\\s(\\d+)', '\\1') \
+      | list }}"
+  when: normal_execution_check is not defined
```
ehelms: I don't see this defined anywhere

obal/data/scratch_build.yml
```diff
@@ -3,17 +3,24 @@
...
         - always
 
-    - include: tasks/tito_release.yml
+    - name: 'Set tito_releasers'
+      set_fact:
+        build_package_tito_releasers: "{{ releasers | join(' ') }}"
+        build_package_tito_args: "{{ build_package_tito_args }} --scratch"
+      when: build_package_koji_command == 'koji'
+      tags:
+        - always
```
ehelms: Based on IRC, this is moving into the role I believe?

pcreech: Yup, hence the WIP

obal/data/release_package.yml
pcreech: Doing so

obal/data/roles/build_package/tasks/wait.yml
pcreech: Yeah, it got removed.   I think i didn't understand at that moment what it actually did, but now that I've played with these paths so much it insured setting of --tags wait when you wanted to wait instead of doing it by default.  Should I put 'normal_execution_check' back instead of requiring '--skip-tags wait' ??

pcreech:
Check build issue created: https://github.com/evgeni/obal/issues/37

PR #38, opened 2018-01-29
Skip diff-brew-package unless otherwise specified

pcreech:
Refactor roles for release_package to be system agnostic

### DIFF
--- a/obal/data/roles/diff_brew_package/tasks/main.yml
+++ b/obal/data/roles/diff_brew_package/tasks/main.yml
@@ -1,121 +1,15 @@
 ---
-- name: get spec file
-  find:
-    pattern: "*.spec"
-    path: "{{ inventory_dir }}/packages/{{ inventory_hostname }}"
-  register: spec_file
+- name: 'Include diff-brew-package'
+  include_tasks: diff.yml
+  when: not diff_brew_package_skip
   tags:
     - always
 
-- name: "Set spec file path, scl options and releasers"
+- name: 'Skip diff-brew-package'
   set_fact:
-    spec_file_path: "{{ spec_file.files[0].path }}"
-    scl_prefix_option_str: "{{ '--define \"scl_prefix tfm-\"' if inventory_hostname.startswith('tfm-') else '' }}"
-    joined_releasers: "{{ releasers | join(' ') }}"
-  tags:
-    - always
-
-- name: "Set branch checks"
-  set_fact:
-    rhel5_branch_exists: "{{ true if 'rhel-5' in joined_releasers else false }}"
-    rhel6_branch_exists: "{{ true if 'rhel-6' in joined_releasers else false }}"
-    rhel7_branch_exists: "{{ true if 'rhel-7' in joined_releasers else false }}"
-  tags:
-    - always
-
-- name: 'Look up current version of package in brew for EL5'
-  shell: "brew list-tagged --quiet \
-    satellite-{{ satellite_version }}-rhel-5-candidate \
-    {{ inventory_hostname }} \
-    | sort --reverse --key 1 --version-sort \
-    | head -n1 \
-    | cut -d' ' -f1"
-  register: latest_brewed_rhel5
-  changed_when: False
-  when: rhel5_branch_exists | bool
-  tags:
-    - always
-
-- name: 'Look up current version of package in brew for EL6'
-  shell: "brew list-tagged --quiet \
-    satellite-{{ satellite_version }}-rhel-6-candidate \
-    {{ inventory_hostname }} \
-    | sort --reverse --key 1 --version-sort \
-    | head -n1 \
-    | cut -d' ' -f1"
-  register: latest_brewed_rhel6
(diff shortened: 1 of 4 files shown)


ehelms:
Could we switch this to move away from brew language to be based on configured build system? This should work for brew or koji.

pcreech:
That's the overall intent (to make it brew/koji agnostic... ) but I wanted to make sure other choices I made so far looked good at this stage.

obal/data/roles/diff_brew_package/tasks/dist.yml
```diff
@@ -0,0 +1,44 @@
+- name: "Set branch checks"
+  set_fact:
+    "{{ item.name }}_branch_exists": "{{ true if item.releaser in releasers else false }}"
```
ekohl: I don't know if the latter is pure python but in pure python ```item.releaser in releasers``` already returns a boolean.

obal/data/roles/diff_brew_package/defaults/main.yml
```diff
@@ -0,0 +1,38 @@
+---
+diff_brew_pkg_dists:
+  - name: rhel5
```
ekohl: Is this here for example purposes? Obal shouldn't carry specifics

obal/data/roles/diff_brew_package/tasks/dist.yml
```diff
@@ -0,0 +1,44 @@
...
+  register: "latest_brewed_{{ item.name }}"
+  changed_when: False
+  when: "hostvars[inventory_hostname][item.name + '_branch_exists'] | bool"
+  tags:
+    - always
+
+- name: 'Calculate {{ item.dist }} RPM version in local specfile'
+  command: "rpmspec --query '{{ spec_file_path }}' \
+    --define 'dist {{ item.dist }}' \
+    --define 'foremandist .{{ foremandist }}' \
```
ekohl: This is rather specific to Foreman. Maybe a more generic approach is needed.

obal/data/roles/diff_brew_package/tasks/dist.yml
```diff
@@ -0,0 +1,44 @@
...
+  set_fact:
+    "{{ item.name }}_branch_exists": "{{ true if item.releaser in releasers else false }}"
+  tags:
+    - always
+
+- name: 'Look up current version of package in {{ build_package_koji_command }} for {{ item.dist }}'
+  shell: "{{ build_package_koji_command }} list-tagged --quiet \
+    {{ item.koji_tag }} \
+    {{ inventory_hostname }} \
+    | sort --reverse --key 1 --version-sort \
```
ekohl: Neat! Something like this would have made my life a lot easier today.

obal/data/roles/diff_brew_package/defaults/main.yml
pcreech: Yes, this data is only there to drive the current brew case while the refactoring is happening

obal/data/roles/diff_brew_package/tasks/dist.yml
pcreech: Yeah, artifact of quick copy-past-edit'in.  Latest push JUST got the python filter working

obal/data/roles/diff_brew_package/tasks/dist.yml
pcreech: Yup, sure is.  Still working on tackling it all

pcreech:
So, went a different route with this.  For now, we are just skipping diff-brew-package unless otherwise specified until a better solution is found

obal/data/roles/diff_brew_package/tasks/main.yml
```diff
@@ -1,121 +1,15 @@
...
-  find:
-    pattern: "*.spec"
-    path: "{{ inventory_dir }}/packages/{{ inventory_hostname }}"
-  register: spec_file
+- name: 'Include diff-brew-package'
+  include_tasks: diff.yml
+  when: not diff_brew_package_skip
   tags:
-    - always
+    always
```
evgeni: can you make this `- always` again? technically both work, but we use the explicit dashed form everywhere else

obal/data/roles/diff_brew_package/tasks/main.yml
```diff
@@ -1,121 +1,15 @@
...
-    - always
-
-- name: "Compared local and brew package versions"
-  set_fact:
-    rhel5_build_needed: "{{ latest_brewed_rhel5 != latest_local_rhel5 }}"
-    rhel6_build_needed: "{{ latest_brewed_rhel6 != latest_local_rhel6 }}"
-    rhel7_build_needed: "{{ latest_brewed_rhel7 != latest_local_rhel7 }}"
-  tags:
-    - always
+    always
```
evgeni: can you make this `- always` again? technically both work, but we use the explicit dashed form everywhere else

pcreech:
fixed the '- always' tags, merging.

PR #39, opened 2018-01-31
add functional tests for obal

evgeni:
This is still mostly WIP, but I wanted to see what y'all are thinking.

The idea is to have a dummy repository and execute the various tasks we have on it.

# current gotchas

* It sets `ansible_remote_tmp` for Travis, as otherwise this fails there :(
* only downstream and upstream repos available, not pulp
* only `scratch`, `release`, 'add' and `update` are implemented at the moment (and the last two only for downstream repos, as the playbooks don't work for upstream ones yet)

Fixes: #36

### DIFF
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -1,0 +1,189 @@
+import functools
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import pytest
+import obal
+
+
+FIXTURE_DIR = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)),
+    'fixtures',
+)
+
+TESTREPO_DIR = os.path.join(FIXTURE_DIR, 'testrepo')
+MOCKBIN_DIR = os.path.join(FIXTURE_DIR, 'mockbin')
+
+DEFAULT_ARGS = []
+if os.environ.get('TRAVIS', None):
+    DEFAULT_ARGS.extend(['-e', 'ansible_remote_tmp=/tmp/ansible-remote'])
+
+
+def obal_cli_test(func=None, repotype='upstream'):
+    if func is None:
+        return functools.partial(obal_cli_test, repotype=repotype)
+
+    @functools.wraps(func)
+    def func_wrapper(*args, **kwargs):
+        tempdir = tempfile.mkdtemp()
+        repodir = os.path.join(tempdir, 'repo')
+        shutil.copytree(TESTREPO_DIR, repodir)
+
+        oldcwd = os.getcwd()
+        oldpath = os.environ['PATH']
+        os.chdir(os.path.join(repodir, repotype))
+        os.environ['PATH'] = "{}:{}".format(MOCKBIN_DIR, oldpath)
+        os.environ['MOCKBIN_LOG'] = os.path.join(tempdir, 'mockbin.log')
+
+        subprocess.check_call(['git', 'init'])
+
+        func(*args, **kwargs)
+
+        os.chdir(oldcwd)
+        os.environ['PATH'] = oldpath
+        os.environ.pop('MOCKBIN_LOG')
+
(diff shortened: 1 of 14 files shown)


tests/test_functional.py
```diff
@@ -0,0 +1,92 @@
...
+TESTREPO_DIR = os.path.join(FIXTURE_DIR, 'testrepo')
+MOCKBIN_DIR = os.path.join(FIXTURE_DIR, 'mockbin')
+
+DEFAULT_ARGS = []
+if os.environ.get('TRAVIS', None):
+    DEFAULT_ARGS.extend(['-e', 'ansible_remote_tmp=/tmp/ansible-remote'])
+
+
+def obal_cli_test(func):
+    def func_wrapper(*args, **kwargs):
```
ekohl: Should this use ```contextlib.contextmanager```?

tests/test_functional.py
```diff
@@ -0,0 +1,92 @@
...
+        tempdir = tempfile.mkdtemp()
```
ekohl: pytest has a [tmpdir fixture](https://docs.pytest.org/en/latest/tmpdir.html). Would that work here as well?

tests/test_functional.py
evgeni: and then the tests would use smth like `with obal_cli_test():`?

evgeni:
and seems the `--tags` stuff does not get tested properly.

tests/test_functional.py
ekohl: Sorry, I meant ```functools.wraps```

evgeni: That makes more sense, implemented. :)

tests/test_functional.py
```diff
@@ -0,0 +1,200 @@
...
+        "['{bin}/tito', 'release', '--scratch', 'dist-git', '-y']"
+    ]
+    assert_mockbin_log(expected_log)
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_release_upstream_hello():
+    sys.argv = ['obal', 'release', 'hello'] + DEFAULT_ARGS
+    with pytest.raises(SystemExit) as excinfo:
+        obal.main()
```
ehelms: The lines 98-100 here appear to be a pattern throughout the tests. Something we could DRY up without harming the readability of the tests?

ehelms: Would the assert need to be in the with block since it references `excinfo` ?

I could see it being nice from a testing perspective to either:

 1) Split the running of the command from the assertion of success or failure of the command
 2) Have two methods that call similar code but expect success or failure

I am thinking of it as being able to run:

```
assert_obal_success 'release hello'
assert_obal_failure 'release blah-blah not real command'
```

My reasoning is when looking at testing, I want to know I am performing an assertion vs. running helper code. This I find helps me with readability of tests.

evgeni: I am not sure. Will test it out.

evgeni: Tried to drain the swamp a bit, added `assert_obal_success` and `assert_obal_failure` wrappers, based on the above `run_obal`. And yes, the `excinfo` assertion has to happen in the same func as the `with`.

tests/test_functional.py
```diff
@@ -0,0 +1,188 @@
...
+    expected_log = "\n".join(content)
+    expected_log = expected_log.replace('{bin}', MOCKBIN_DIR)
+    with open(os.environ['MOCKBIN_LOG']) as mockbinlog:
+        log = mockbinlog.read().strip()
+        assert log == expected_log
+
+
+def setup_upstream(upstream_path):
+    # create a cloneable upstream repo, what we ship in fixtures is not
+    subprocess.call(['git', 'init'], cwd=upstream_path)
```
ekohl: Would ```check_call``` be safer?

evgeni: good catch, done

PR #40, opened 2018-02-05
Get obal to work with pulp's nightly process

pcreech:
Get obal to work with pulp's nightly process

### DIFF
--- a/obal/data/roles/setup_sources/tasks/git.yml
+++ b/obal/data/roles/setup_sources/tasks/git.yml
@@ -1,0 +1,17 @@
+---
+- name: 'Set source checkout location'
+  set_fact:
+    setup_sources_git_checkout_dir: "/tmp/ansible_git_{{ inventory_hostname }}"
+
+- name: 'Clone latest git'
+  git:
+    repo: "{{ setup_sources_git }}"
+    dest: "{{ setup_sources_git_checkout_dir }}"
+    version: "{{ git_branch|default('master') }}"
+    force: yes
+
+- name: 'Register source dir tito argument'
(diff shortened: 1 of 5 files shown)


obal/data/roles/setup_sources/tasks/git.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: 'Set source checkout location'
+  set_fact:
+    setup_sources_git_checkout_dir: "/tmp/ansible_git_{{ inventory_hostname }}"
+
+- name: 'Ensure git location'
+  file:
```
ehelms: Does the git module not create the directory by default?

obal/data/roles/setup_sources/tasks/main.yml
```diff
@@ -5,28 +5,14 @@
...
     - always
 
-- name: get spec file
-  find:
-    pattern: "*.spec"
-    path: "{{ package_dir }}"
-  register: spec_file
+- name: 'Setup sources from specfile'
+  include_tasks: specfile.yml
+  when: git is not defined
```
ehelms: Should this be `setup_sources_git` ?

obal/data/roles/setup_sources/tasks/git.yml
pcreech: yup, this was a relic, removed

obal/data/roles/setup_sources/tasks/main.yml
pcreech: Yup, relic from testing with pulp's current nightly code.  fixed

PR #45, opened 2018-02-13
Add git annex spec builder for tito

ehelms:
This would allow it to be used across repositories without having to contribute this big of a change to tito.

### DIFF
--- a/obal/tito_extensions/git_annex_spec_builder.py
+++ b/obal/tito_extensions/git_annex_spec_builder.py
@@ -1,0 +1,153 @@
+import os
+import shutil
+from distutils.version import LooseVersion as loose_version
+from pkg_resources import require
+
+from tito.compat import getstatusoutput
+from tito.config_object import ConfigObject
+from tito.builder import GitAnnexBuilder
+from tito.builder.main import BuilderBase
+from tito.common import error_out, debug, run_command, get_spec_version_and_release, \
+    find_spec_like_file, get_relative_project_dir_cwd, warn_out
+
+
+class GitAnnexSpecBuilder(GitAnnexBuilder):
+
+    def __init__(self, name=None, tag=None, build_dir=None,
+                 config=None, user_config=None,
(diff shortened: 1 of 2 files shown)


evgeni:
`flake8` is angry at your code ;)

I wonder if we should put that into `obal.tito_extensions` or something similar as a namespace, just to make double sure there can't be multiple `tito` definitions?

ehelms:
`flake8` is happy again

ekohl:
Does this mean the setup_sources script would no longer be needed for tito builds?

ehelms:
No... this is what will allow us to drop the use of git tags in the packaging repository

obal/tito_extensions/git_annex_spec_builder.py
```diff
@@ -0,0 +1,153 @@
...
+                    self.config.get("requirements", "tito"),
+                    "Your version: %s" % require('tito')[0].version
+                ])
+
+        self.display_version = self._get_display_version()
+
+        self.relative_project_dir = get_relative_project_dir_cwd(self.git_root)
+
+        tgz_base = self._get_tgz_name_and_ver()
+        self.tgz_filename = tgz_base + ".tar.gz"
```
ekohl: Would this fail on [nodejs-babel-cli](https://github.com/theforeman/foreman-packaging/tree/8446d10183de7af1bf73ca100ff181684f3b48fd/nodejs-babel-cli) since the extension is .tgz? It also has multiple source file.

ehelms: Good question, but this tarball is actually generated by tito, it wraps up the spec and sources into a tarball for it's inner workings so this shouldn't break any one package.

evgeni:
given this is what we have been using w/o any issues downstream for quite some time now, I'd lean to merge it as-is and fix up any corner cases as we go on.

:+1: 

@ekohl fine with you?

PR #48, opened 2018-02-13
Add playbook to cleanup copr scratch repos

ehelms:


obal/data/cleanup_copr.yml
```diff
@@ -0,0 +1,17 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: 'Get Copr repositories'
+      command: "copr-cli list"
```
evgeni: should this be an explicit `copr-cli list @theforeman`?

obal/data/cleanup_copr.yml
```diff
@@ -0,0 +1,17 @@
...
+      register: copr_repos
+
+    - set_fact:
+        scratch_repos: "{{ copr_repos.stdout.split('\n') | select('match', '^Name:') | list | replace('Name: ', '') }}"
```
evgeni: the fact should be `copr_scratch_repos`, right?

obal/data/cleanup_copr.yml
```diff
@@ -0,0 +1,17 @@
...
+
+    - set_fact:
+        scratch_repos: "{{ copr_scratch_repos | select('match', '.+scratch.+') | list }}"
```
evgeni: the fact should be `copr_scratch_repos`, right?

evgeni:
do we want a `obal cleanup` action for that? or is that very copr specific and you'd use `ansible-playbook` for that?

ehelms:
I added an action and updated the variables. This cleanup is currently copr specific. I added 'cleanup-copr' but I wonder if `copr cleanup` should be the command?

evgeni:
`copr cleanup` is something `obal` would not understand today, as it expects the action to be a single world and all following to be package names. so I'd stick with the current name for now.

`ansible-lint` is unhappy, but LGTM otherwise

ehelms:
Updated to fix the linting errors

PR #49, opened 2018-02-15
Allow custom package directories

pcreech:
Allow custom package directories


pcreech:
Bleh, VSCode strikes again!  Fixing newlines real fast

pcreech:
Newlines added

PR #50, opened 2018-02-17
Re-factor package diffing to simplify and support copr

ehelms:
This is a fairly large change in workflow to add copr support and simplify the brew package diff. This will require any users to define more information in their `package_manifest.yml`.

For copr:

```
+    scl: tfm-ror51
+    copr_user: "@theforeman"
+    diff_package_copr_repo: tfm-ror51-staging
+    diff_package_type: copr
```

For brew:

```
+    scl: tfm
+    brew_tags:
+      - satellite-6.3.0-rhel-7-candidate
```

This changes to requiring more variables to be defined explicitly but has the advantage of simplifying the code and reducing the assumptions being made. No need to parse tito or make assumptions about tito layout. The project defines on a per group or per package basis the brew tags or in copr's case the repository to check against. 

The second big change is this reduces the needs build check to a substring check. This attempts to ignore whatever dist is by calculating the NVR locally and checking that this string is IN the official NVR coming from the build system.


evgeni:
Can you please adjust the shipped test repos to include the required metadata?

obal/data/roles/diff_package/tasks/git_package_info.yml
```diff
@@ -0,0 +1,25 @@
...
+- name: "Set spec file path, scl options and releasers"
+  set_fact:
+    spec_file_path: "{{ spec_file.files[0].path }}"
+
+- name: 'Get latest package info from git'
+  command: >
+    rpmspec --query
+      --undefine=dist
+      --queryformat=%{name}-%{version}-%{release}
+      {{ '--define "scl_prefix ' + scl + '-"' if scl else '' }}
```
evgeni: This currently fails when `scl` is not defined at all. Should the `scl` var be defined as `''` in `defaults/main.yml`?

obal/data/roles/diff_package/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+- block:
```
evgeni: What is this block for?

obal/data/roles/diff_package/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
...
+    - name: 'Include brew package diff'
```
evgeni: Why not

```yaml
   - name: 'Include {{ diff_package_type }} package diff'
     include_tasks: "{{ diff_package_type }}.yml"
     tags:
       - always
```

Maybe with a `when: diff_package_type in ['brew', 'copr']` to allow non-diffed repos?

obal/data/roles/diff_package/tasks/copr.yml
```diff
@@ -0,0 +1,33 @@
...
+- set_fact:
+    copr_package_version: "{{ copr_package_info['name'] + '-' + copr_package_info['latest_build']['pkg_version'] }}"
+  when: copr_package_info | success
+
+- set_fact:
+    copr_package_version: ""
+  when: copr_package_info | failed
+
+- set_fact:
+    diff_package_changed: "{{ git_package_version not in copr_package_version }}"
```
evgeni: Should this be a check for "startswith" (`{{ not copr_package_version.startswith(git_package-version) }}`)?

obal/data/roles/diff_package/tasks/brew.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: 'Look up current version of package in brew'
+  shell: "brew list-tagged --quiet --latest {{ item }} {{ inventory_hostname }} | cut -d' ' -f1"
+  register: brew_package_versions
+  tags:
+    - always
+  with_items: "{{ brew_tags }}"
+
+- set_fact:
+    diff_package_changed: "{{ git_package_version not in item.stdout or diff_package_changed }}"
```
evgeni: Should this be a check for "startswith" (`{{ not item.stdout.startswith(git_package-version) or diff_package_changed }}`)?

obal/data/roles/diff_package/defaults/main.yaml
```diff
@@ -0,0 +1,3 @@
+---
+diff_package_type: brew
+diff_package_skip: true
```
evgeni: This isn't used anymore?

ehelms: I adjusted the -block in main.yml to use this now.

obal/data/roles/diff_package/tasks/main.yml
ehelms: Check the `when` condition at the bottom now.

obal/data/roles/diff_package/tasks/git_package_info.yml
ehelms: Added a default for this.

ehelms:
Generally updated for all comments addressed and a few clean ups that were missing.

obal/data/roles/diff_package/tasks/copr.yml
```diff
@@ -0,0 +1,33 @@
...
+- set_fact:
+    copr_package_version: "{{ copr_package_info['name'] + '-' + copr_package_info['latest_build']['pkg_version'] }}"
+  when: copr_package_info | success
+
+- set_fact:
+    copr_package_version: ""
+  when: copr_package_info | failed
+
+- set_fact:
+    diff_package_changed: "{{ copr_package_version.startswith(git_package_version) }}"
```
evgeni: Shouldn't this be `not copr_package_version.startswith(git_package_version)`?

obal/data/roles/diff_package/tasks/brew.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: 'Look up current version of package in brew'
+  shell: "brew list-tagged --quiet --latest {{ item }} {{ inventory_hostname }} | cut -d' ' -f1"
+  register: brew_package_versions
+  tags:
+    - always
+  with_items: "{{ brew_tags }}"
+
+- set_fact:
+    diff_package_changed: "{{ item.stdout.startswith(git_package_version) or diff_package_changed }}"
```
evgeni: shouldn't this be `not item.stdout.startswith(git_package_version)`

obal/data/roles/diff_package/tasks/main.yml
evgeni: yeah. btw, I think now the block does not  need to be a block, as it only has one task in it.

ehelms:
Updated, thanks for the diff!

obal/data/roles/diff_package/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+- block:
```
evgeni: Given this `block` now only contains one task, should we remove the `block` completely?

tests/fixtures/testrepo/downstream/package_manifest.yaml
```diff
@@ -1,12 +1,15 @@
...
-    diff_brew_package_skip: false
+    diff_package_skip: false
+    scl: tfm
     build_package_koji_command: 'brew'
     satellite_version: '6.3.0'
     foremandist: 'fm1_15'
     releasers:
       - obaltest-dist-git-rhel-7
+    brew_tags:
+      - satellite-6.3.0-rhel-7-candidate
```
evgeni: Should we rename this to `obaltest-rhel-7-candidate`? Just so that we don't have `satellite` in here?

ehelms:
Updated all comments. Lets see if tests are happy.

tests/fixtures/testrepo/downstream/package_manifest.yaml
```diff
@@ -1,12 +1,15 @@
 ---
 packages:
   vars:
-    diff_brew_package_skip: false
+    diff_package_skip: false
+    scl: tfm
     build_package_koji_command: 'brew'
-    satellite_version: '6.3.0'
+    obaltest_version: '6.3.0'
```
evgeni: LOL, thanks for spotting this. This line can go now, after your refactoring. :boom:

tests/fixtures/testrepo/empty/package_manifest.yaml
```diff
@@ -1,9 +1,9 @@
 ---
 packages:
   vars:
-    diff_brew_package_skip: false
+    diff_package_skip: false
     build_package_koji_command: 'brew'
-    satellite_version: '6.3.0'
+    obaltest_version: '6.3.0'
```
evgeni: LOL, thanks for spotting this. This line can go now, after your refactoring. :boom:

ehelms:
Nuked'em

PR #51, opened 2018-02-17
Add repoclosure support

ehelms:
From my testing this works and is fairly simple as its a single playbook, however, I am torn on the design. Other ideas:

 * Move repoclosure into a role to make the variables a bit more clean by defining defaults for them, this would make the playbook simply a call to the role
 * Create a repoclosure module to handle the data manipulation this possibly means you could define multiple repoclosures and use with_items to iterate on them (the current solution is to call repoclosure multiple times with the variables needed for each configuration for each repoclosure that needs computed)

One thing that was a bit unclean about this is that repoclosure does not operate on packages. So the call is currently `obal repoclosure all`. The all being the slightly ugly part.

### DIFF
--- a/obal/data/repoclosure.yml
+++ b/obal/data/repoclosure.yml
@@ -1,0 +1,4 @@
+- hosts: localhost
+  gather_facts: no
+  roles:
+    - repoclosure
--- a/obal/data/roles/repoclosure/tasks/main.yml
+++ b/obal/data/roles/repoclosure/tasks/main.yml
@@ -1,0 +1,27 @@
+---
+- name: 'Build additional repos'
+  set_fact:
+    additional_repos: "{{ additional_repos|default([]) + ['--repofrompath=' + item.name + ',' + item.url] }}"
+  with_items: "{{ repoclosure_additional_repos }}"
+
+- name: 'Run repoclosure'
+  command: >
+    repoclosure
+      -c {{ inventory_dir }}/{{ repoclosure_config }}
+      -t
+      -r {{ repoclosure_target_repo }}
+      {{ additional_repos | join(' ') }}
+      {{ '-l ' if repoclosure_look_aside_repos else '' }}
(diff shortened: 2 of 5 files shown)


obal/data/repoclosure.yml
```diff
@@ -0,0 +1,32 @@
...
+      command: >
+        repoclosure
+          -c {{ inventory_dir }}/{{ repoclosure_config }}
+          -t
+          -r {{ repoclosure_target_repo }}
+          {{ additional_repos | join(' ') }}
+          {{ '-l ' if repoclosure_look_aside_repos else '' }}
+          {{ repoclosure_look_aside_repos | join(' -l ') }}
+      register: output
+      when: output is not defined
```
evgeni: Why is this `when` necessary?

evgeni:
I think I'd go with the role model for now. Simple playbooks are good playbooks ;)

Also I think this warrants an entry in the README, explaining all the different variables you can/must set for this to work.

obal/data/repoclosure.yml
ehelms: The ansible-lint rule not to run operations if there are no changes.

ehelms:
Updated!

obal/data/roles/repoclosure/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: 'Build additional repos'
+  set_fact:
+    additional_repos: "{{ additional_repos|default([]) + ['--repofrompath=' + item.name + ',' + item.url] }}"
+  with_items: "{{ repoclosure_additional_repos }}"
+  when: repoclosure_additional_repos is defined
```
evgeni: Isn't it always defined, as it's in `defaults/main.yml`?

ehelms: Truth. Fixed.

PR #53, opened 2018-02-21
Add __main__.py to allow running python -m obal

ekohl:
It still shows an ugly --help but you should just call it with the correct arguments :)


obal/__main__.py
```diff
@@ -0,0 +1,2 @@
+from __init__ import main
```
evgeni: why not `from obal import main`?

PR #54, opened 2018-02-21
Fix spec builder to work with whitelists

pcreech:


obal/tito_extensions/git_annex_spec_builder.py
```diff
@@ -84,13 +84,20 @@ def __init__(self, name=None, tag=None, build_dir=None,
...
+        Figure out the git tag and version-release we're building.
+        """
+        # Determine which package version we should build:
+        build_version = None
+        if self.build_tag:
+            build_version = self.build_tag[len(self.project_name + "-"):]
+        else:
+            build_version = get_spec_version_and_release(self.start_dir,
+                                                         find_spec_like_file(self.start_dir))
+            self.build_tag = self._get_tag_for_version(build_version)
```
ehelms: Do we want this here? Will that try to look up a git tag?

pcreech: It's what helps set the build tag properly, and no it will not look up a git tag:

https://github.com/dgoodwin/tito/blob/master/src/tito/builder/main.py#L471

PR #59, opened 2018-02-27
[WIP] Add obal-fmt

ekohl:
This is a very simple tool that formats the YAML manifest and can be used to enforce a style.

### DIFF
--- a/obal/__init__.py
+++ b/obal/__init__.py
@@ -19,6 +19,8 @@ from functools import total_ordering
 import yaml
 from pkg_resources import resource_filename
 
+from obal import formatting
+
 try:
     import argcomplete
 except ImportError:
@@ -164,6 +166,13 @@ class Playbook(object):
         return NotImplemented
 
 
+def get_manifest_path():
+    """
+    Return the package manifest path
+    """
+    return os.path.join(os.getcwd(), 'package_manifest.yaml')
+
+
 def find_playbooks(playbooks_path):
     """
     Find all playbooks in the given path.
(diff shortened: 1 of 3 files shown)


ehelms:
I am commenting on the usage in f-packaging as I think its easier to see there.

ekohl:
> I am commenting on the usage in f-packaging as I think its easier to see there.

That was exactly the reason why I opened that PR.

obal/__init__.py
```diff
@@ -7,6 +7,7 @@
 import os
 import sys
 
+import yaml
```
evgeni: Strictly speaking, this needs `PyYAML` as a dep in `setup.py` etc, right? (it gets pulled in by `ansible` anyways, but let's be explicit)

obal/__init__.py
```diff
@@ -105,6 +106,33 @@ def generate_ansible_args(inventory_path, playbook_path, args):
...
+        argcomplete.autocomplete(parser)
+
+    args = parser.parse_args(cliargs)
+
+    if not os.path.exists(args.manifest):
+        print("Could not find your package_manifest.yaml")
+        exit(1)
+
+    with open(args.manifest) as manifest_fp:
+        manifest = yaml.load(manifest_fp)
```
evgeni: should that be `safe_load`?

ekohl:
Updated to maintain ordering when ```packages``` is a dictionary key. I would like this to be more explicit rather than respecting current ordering.

ekohl:
I want to look into compatibility with ruamel 0.13 since that's in Fedora 28 and ruamel 0.15 since that's on PyPi and in F29.

evgeni:
closed in favor of the obsah issue

PR #61, opened 2018-02-28
Use array for tito releaser args

ehelms:


ehelms:
This will allow defining as in https://github.com/theforeman/foreman-packaging/pull/2235#issuecomment-369299400

ekohl:
How does this deal with strings? In python calling join on a string leads to ... interesting results.

ehelms:
I don't know actually. If a user passes a string, then this will likely error and thats on the user. Same would be true of any parameter that was improperly sent through in Ansible I think.

evgeni:
I could not find any users of this at the moment, so I am fine with "breaking" it.

@ekohl OK for ya?

PR #62, opened 2018-02-28
Fix copr scratch builds

ehelms:


evgeni:
I wonder how #60 passed tests, our mocked `tito` does not know what `--test` is :)

Issue #65, opened 2018-03-02
Stop using `always`


evgeni:
It's 3 am and I'm thinking about ansible tags, halp.

We're currently tagging all of our tasks with `always` so that they are still executed when we run `--tags wait` or similar. This is super cumbersome and errorprone, so let's not do it. Instead we can just run `--tags untagged` by default, and `untagged,wait` when we want to wait. Should also get rid of the `normal_execution_check`.

Docs: http://docs.ansible.com/ansible/latest/playbooks_tags.html#special-tags

PS: backticks are totally hidden on the Android keyboard :(

PR #67, opened 2018-03-02
stop using tags to limit what we run

evgeni:
instead use two boolean vars to tell the build_package role whether it should wait and download logs.

also changes the default to wait, users can opt out by passing `-e build_package_wait=False`

Closes: #65

### DIFF
--- a/obal/data/roles/build_package/tasks/copr.yml
+++ b/obal/data/roles/build_package/tasks/copr.yml
@@ -3,25 +3,17 @@
   set_fact:
     build_package_tito_args: "{{ build_package_tito_args }} --test"
   when: build_package_test|default(false)
-  tags:
-    - always
 
 - block:
     - name: 'Set tito_releasers for copr release'
       set_fact:
         build_package_tito_releasers: "{{ releasers | join(' ') }}"
-      tags:
-        - always
 
     - include_tasks: tito_release.yml
-      tags:
-        - always
 
     - name: 'Wait for tasks to finish'
       include_tasks: wait.yml
-      when: build_package_normal_execution_check is not defined
-      tags:
-        - wait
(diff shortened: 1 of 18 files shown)


evgeni:
@komidore64 would love your thoughts on this, as the original implementation was yours

komidore64:
the main reason for the `always` tags was so that you had bunches of stuff that'd run without passing tags, but also run the same tasks **plus** tagged tasks, so i'm okay with this.

now since these playbooks are wrapped by obal, we're not running ansible directly, so i'm okay with them being removed. just make sure it works otherwise **i'm coming after you**.

ehelms:
Could we consider making the wait functionality the default? Given most jobs need to wait.

komidore64:
I support this.

komidore64:
@evgeni 

make it so!

obal/data/roles/build_package/tasks/copr.yml
```diff
@@ -49,25 +37,17 @@
...
       command: 'tito build --srpm --scl={{ scl }}'
       args:
         chdir: "{{ inventory_dir }}/packages/{{ inventory_hostname }}"
       register: srpm_build
-      tags:
-        - always
 
     - name: 'Run build'
-      command: "copr-cli build {{ copr_repo_name }} {{ srpm_build.stdout.split('Wrote: ')[1] }}"
+      command: "copr-cli build {% if not build_package_wait|bool %}--nowait{% endif %} {{ copr_repo_name }} {{ srpm_build.stdout.split('Wrote: ')[1] }}"
```
ehelms: I don't think I realized you could use this style of Jinja2 templating

komidore64:
ack

PR #68, opened 2018-03-02
Make update_package handle base scenario of bumping source by a version

ehelms:


ehelms:
This could likely use some re-factoring to clean it up and break pieces down. I was not sure what roles to break this into but wanted to make it usable. I generated https://github.com/theforeman/foreman-packaging/pull/2245 from this.

obal/data/update_package.yml
```diff
@@ -5,12 +5,15 @@
...
       fail:
-        msg: "'upstream_files' not defined for {{ inventory_hostname }}. Please define within package_manifest.yml"
-      when:
-        - upstream_files is not defined
-        - downstream_version is not defined
+        msg: "version must be specified"
+      when: version is not defined
+
+    - name: 'Create bit branch'
+      command: "git checkout -b {{ inventory_hostname }}-{{ version }}"
```
ekohl: For foreman-packaging I like to prepend ```rpm/``` or ```deb/```. Maybe a prefix parameter could be added?

obal/data/update_package.yml
```diff
@@ -47,3 +50,8 @@
       args:
         chdir: "{{ inventory_dir }}"
       when: new_source is defined
+
+    - name: 'Commit update'
+      command: "git commit -a -m 'Release {{ inventory_hostname }} {{ version }}'"
```
ekohl: In foreman-packaging I don't use ```-a``` when updating but explicitly add the package instead. Sometimes I have some tooling changes open and then it's nice not to accidentally commit those.

obal/data/tasks/download_and_extract_source.yml
```diff
@@ -114,26 +42,31 @@
...
         path: "{{ spec_file_path }}"
         regexp: '^(?P<prefix>{{ item }}\s+)[0-9.]+(?P<suffix>.*)$'
-        replace: '\g<prefix>{{ downstream_release }}\g<suffix>'
+        replace: '\g<prefix>{{ release }}\g<suffix>'
       with_items:
         - 'Release:'
         - '%global _release'
         - '%global release'
 
+    - name: 'Extract git user'
```
ekohl: It may be easier to rely on ```rpmdev-packager```:
```
$ rpmdev-packager 
Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl>
```

obal/data/tasks/download_and_extract_source.yml
```diff
@@ -8,68 +8,6 @@
...
-  when: #block
-    - upstream.endswith('src.rpm')
-    - downstream_version is not defined
-  rescue: #block
-    - name: 'Remove package directory'
-      file:
-        state: 'absent'
-        path: "./{{ package_base_dir }}/{{ inventory_hostname }}"
-
 - block:
```
ekohl: I don't have experience with ```rpmdev-bumpspec``` but that's supposed to do pretty much the same: ```rpmdev-bumpspec --comment 'Release {{ inventory_hostname }} {{ version }}' --new '{{ version }}' {{ spec_file }}```

On the other hand, here you also try to replace globals which rpmdev-bumpspec can only do with release but not versions.

obal/data/update_package.yml
```diff
@@ -5,12 +5,19 @@
...
-        - downstream_version is not defined
+        msg: "version must be specified"
+      when: version is not defined
+
+    - name: 'Git current branch'
+      command: 'git rev-parse --abbrev-ref HEAD'
+      register: git_branch
+
+    - name: 'Create bit branch'
+      command: "git checkout -b {{ git_branch.stdout }}-{{ inventory_hostname }}-{{ version }}"
```
ehelms: @ekohl I modified this to get the current branch and prepend that. So in theory someone might get:

```
master-tfm-ror51-rubygem-sinatra
rpm/develop-rubygem-foreman-tasks
master-pulp
```

Thoughts?

obal/data/tasks/download_and_extract_source.yml
ehelms: Yea, we have some specs that go outside the norm using globals :/

obal/data/update_package.yml
ekohl: @ehelms I meant a parameter the user could provide on the command line (or maybe read from the package manifest)?

obal/data/tasks/download_and_extract_source.yml
```diff
@@ -114,26 +42,27 @@
...
+      command: "rpmdev-packager"
+      register: packager_info
+
     - name: 'add changelog entry'
       lineinfile:
         path: "{{ spec_file_path }}"
         line: |
-          * {{ lookup('pipe', 'date +"%a %b %d %Y"') }} {{ downstream_changelog_name | default('Satellite6 Jenkins') }} {{ downstream_version }}-{{ downstream_release }}
-          {{ downstream_changelog }}
+          * {{ lookup('pipe', 'date +"%a %b %d %Y"') }} {{ packager_info.stdout }} {{ version }}-{{ release }}
```
ekohl: For those of us with different locales it might be good to explicitly set ```$LANG```

obal/data/update_package.yml
```diff
@@ -5,12 +5,19 @@
...
       fail:
-        msg: "'upstream_files' not defined for {{ inventory_hostname }}. Please define within package_manifest.yml"
-      when:
-        - upstream_files is not defined
-        - downstream_version is not defined
+        msg: "version must be specified"
+      when: version is not defined
+
+    - name: 'Git current branch'
+      command: 'git rev-parse --abbrev-ref HEAD'
```
evgeni: needs `chdir` param :)

evgeni:
this needs a rebase after my code shuffling, sorry

evgeni:
so I've been thinking about this a bit, esp on "how do we define `update`?"

When thinking in upstream (`foreman-packaging`, `tfm-ror51`, `pulp-packaging`) terms, the only sensible update we can do is to a newer version of the package, which boils down to "update `%version`, bump `%changelog`, fetch and add the new source" (= what you have implemented here).

For downstream, this is more granular. It can be the same as above (but pulling in a patched version from our internal source server). Or it can be "get me whatever foreman upstream has" (or EPEL, when we slurp in a `src.rpm`).

I think the "bump spec & fetch source" case is the most important (as it saves the most manual work), but I think I'd prefer not to lose the "fetch from foreman" case either, as it also saved me a lot of time when preparing downstream builds.

@ehelms do you think we should keep support for the downstream usecases? and should that be done in this PR or in a separate one?

ekohl:
I think we need an update strategy. Some options I can think of:
* `bump` increase version, pull the source and add it
* `gem` pull the latest from rubygems (I intend to expand https://github.com/theforeman/foreman-packaging/blob/rpm/develop/bump_rpm.sh with updating the dependencies similar to `gem2rpm`)
* `npm` pull the latest from npm (what https://github.com/theforeman/foreman-packaging/blob/rpm/develop/add_npm_package.sh already does)
* `pypi` pull the latest from pypi
* `upstream` what we currently do when puling to downstream

evgeni:
yeah, having a `strategy` setting sounds good to me.

obal/data/tasks/download_and_extract_source.yml
ehelms: What will that affect?

ekohl:
```
$ LANG=nl_NL.UTF-8 date +"%a %b %d %Y"
do mrt 22 2018
```

obal/data/playbooks/update.yml
```diff
@@ -55,3 +48,13 @@
...
       when:
         - "'://' not in item"
+
+    - name: 'Create git branch'
+      command: "git checkout -b {{ inventory_hostname }}-{{ version }}"
+      args:
+        chdir: "{{ inventory_dir }}"
+
+    - name: 'Create git commit'
+      command: "git commit -a -m 'Add {{ inventory_hostname }} {{ version }}'"
```
evgeni: `s/Add/Update/`

obal/data/roles/fetch_sources/tasks/main.yml
```diff
@@ -112,22 +117,12 @@
...
         - '%global release'
 
-    - name: 'add changelog entry'
-      lineinfile:
-        path: "{{ spec_file_path }}"
-        line: |
-          * {{ lookup('pipe', 'date +"%a %b %d %Y"') }} {{ downstream_changelog_name | default('Satellite6 Jenkins') }} {{ downstream_version }}-{{ downstream_release }}
-          {{ downstream_changelog }}
-        insertafter: '^%changelog'
+    - name: 'ensure changelog'
```
evgeni: `ensure changelog` sounds weird to me. the `changelog` module always (well, as long there is a `%changelog` section) adds the entry, so `add changelog entry` as the name here would be more appropriate?

obal/data/roles/fetch_sources/tasks/main.yml
```diff
@@ -112,22 +117,12 @@
...
+      changelog:
+        spec: "{{ spec_file_path }}"
+        changelog: "{{ changelog }}"
+      when: changelog is defined
```
evgeni: `changelog` is always defined, as we have a `set_fact` above, so maybe this `when` is not needed?

obal/data/roles/fetch_sources/tasks/main.yml
```diff
@@ -82,25 +84,28 @@
 
     - name: 'Set default downstream changelog'
```
evgeni: it's now downstream specific anymore, right?

obal/data/roles/fetch_sources/tasks/main.yml
```diff
@@ -82,25 +84,28 @@
...
         regexp: '^(?P<prefix>{{ item }}\s+)[0-9.]+(?P<suffix>.*)$'
-        replace: '\g<prefix>{{ downstream_version }}\g<suffix>'
+        replace: '\g<prefix>{{ version }}\g<suffix>'
       with_items:
         - 'Version:'
         - '%global _version'
         - '%global mainver'
+      when: version is defined
 
     - name: 'Bump release in specfile'
```
evgeni: should this drop the `downstream_` prefix too?

tests/test_functional.py
```diff
@@ -276,7 +276,7 @@ def test_obal_release_downstream_hello_wait_download_rpms():
 def test_obal_update_downstream_hello():
     setup_upstream('../upstream/')
 
-    assert_obal_success(['update', 'hello'])
+    assert_obal_success(['update', 'hello', '-e', 'version=2.10', '-e', 'downstream=true'])
```
evgeni: why is `downstream=true` needed? I can't find such a variable anywhere in our codebase.

also, I don't think an downstream update should require a version, as that scenario should be "just copy over stuff", no?

evgeni: oooh, but you use the version in the commit msg, fair enough.

evgeni:
oh, and this needs an update of the readme :)

tests/test_functional.py
evgeni: passes localy w/o `downstream=true`

obal/data/roles/fetch_sources/tasks/main.yml
```diff
@@ -112,22 +117,12 @@
...
+    - name: 'ensure changelog'
+      changelog:
+        spec: "{{ spec_file_path }}"
+        changelog: "{{ changelog }}"
+      when: changelog is defined
 
     - name: 'update source files'
       include_role:
         name: setup_sources
-
```
evgeni: this means we *always* add a changelog entry, which I think we should *not* do in the case of a downstream "copy from upstream" scenario.

obal/data/roles/fetch_sources/tasks/main.yml
ehelms: Double check the `when` conditions on the set_fact above. This is not always defined. For example, when copying a package from an upstream git repo to a downstream git repo. Our tests are what caught this little nugget.

evgeni: ah, right!

obal/data/roles/fetch_sources/tasks/main.yml
ehelms: I think this is addressed by -- https://github.com/theforeman/obal/pull/68#discussion_r215586801

evgeni: yes it is

obal/data/playbooks/update.yml
```diff
@@ -55,3 +48,13 @@
       with_items: "{{ new_source_urls.stdout_lines | list }}"
       when:
         - "'://' not in item"
+
+    - name: 'Create git branch'
+      command: "git checkout -b {{ inventory_hostname }}-{{ version }}"
```
evgeni: this makes `version` mandatory, but `obal/data/roles/fetch_sources/tasks/main.yml` uses `version` to differentiate between "plain update" (`version is defined`) and "copy from upstream" (`version is not defined`)

ehelms:
@evgeni I updated based on your comments. Let me know what you think.

obal/data/roles/fetch_sources/tasks/main.yml
```diff
@@ -74,60 +74,53 @@
...
-        - downstream_changelog is not defined
+        - changelog is not defined
+        - version is defined
 
     - name: 'Ensure Source0 specfile points to source server'
       replace:
         path: "{{ spec_file_path }}"
         regexp: '^(?P<prefix>Source0:\s+)http.*/(?P<filename>.*)$'
         replace: '\g<prefix>{{ source_server }}\g<filename>'
+      when: source_server is defined
```
evgeni: `source_server` tends to be defined on a group/global level, so this changes the `Source0` line even on "fetch from upstream" scenarios. Should this test for `source_server` *and* `version` being defined?

ehelms: The implication being if version and source_server are defined then assume that strategy? Would you ever do one without the other? Sounds like we may need to re-think this in general around defined strategies. For now, it cna help avoid issues so I'll update.

ehelms:
Updated

evgeni:
I have extended the tests with the following patch, and they still pass :+1: 

```patch
diff --git a/tests/fixtures/testrepo/downstream/packages/hello/hello.spec b/tests/fixtures/testrepo/downstream/packages/hello/hello.spec
index b23da0a..a4eb2c1 100644
--- a/tests/fixtures/testrepo/downstream/packages/hello/hello.spec
+++ b/tests/fixtures/testrepo/downstream/packages/hello/hello.spec
@@ -47,5 +47,5 @@ fi
 %license COPYING
 
 %changelog
-* Tue Sep 06 2011 The Coon of Ty <Ty@coon.org> 2.10-1
+* Tue Sep 06 2011 The Coon of Ty <Ty@coon.org> 2.9-1
 - Initial version of the package
diff --git a/tests/test_functional.py b/tests/test_functional.py
index c529bb3..61d2d36 100644
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -272,6 +272,21 @@ def test_obal_release_downstream_hello_wait_download_rpms():
 
 
+def test_obal_update_upstream_hello():
+    assert_obal_success(['update', 'hello', '-e', 'version=2.8'])
+
+    assert not os.path.exists('packages/hello/hello-2.10.tar.gz')
+    assert not os.path.islink('packages/hello/hello-2.10.tar.gz')
+    assert os.path.islink('packages/hello/hello-2.8.tar.gz')
+
+    with open('packages/hello/hello.spec') as specfile:
+        specfilecontent = specfile.read()
+
+    assert 'Version:        2.8' in specfilecontent
+    assert '- Release hello 2.8' in specfilecontent
+
+
 @obal_cli_test(repotype='downstream')
 def test_obal_update_downstream_hello():
     setup_upstream('../upstream/')
@@ -282,12 +297,22 @@ def test_obal_update_downstream_hello():
     assert not os.path.islink('packages/hello/hello-2.9.tar.gz')
     assert os.path.islink('packages/hello/hello-2.10.tar.gz')
 
+    with open('packages/hello/hello.spec') as specfile:
+        specfilecontent = specfile.read()
+
+    # a downstream update overwrites all the data with whatever is in upstream
+    # thus we expect to see no mention of 2.9 in the specfile anymore
+    # we also don't expect a changelog entry added
+    assert 'Version:        2.10' in specfilecontent
+    assert '2.9' not in specfilecontent
+    assert '- Release hello' not in specfilecontent
+
 
 @obal_cli_test(repotype='empty')
 def test_obal_add_downstream_hello():
     setup_upstream('../upstream/')
 
-    assert_obal_success(['add', 'hello', '-e', 'downstream=true'])
+    assert_obal_success(['add', 'hello'])
 
     assert os.path.islink('packages/hello/hello-2.10.tar.gz')
 
```

obal/data/playbooks/update.yml
```diff
@@ -55,3 +48,27 @@
...
+    - name: "Set spec file path, scl options and releasers"
+      set_fact:
+        spec_file_path: "{{ spec_file.files[0].path }}"
+
+    - name: 'Get updated version from spec'
+      command: "rpmspec --query --queryformat=%{version} --srpm {{ spec_file_path }}"
+      register: updated_version
+
+    - name: 'Create git branch'
+      command: "git checkout -b {{ inventory_hostname }}-{{ updated_version.stdout }}"
```
evgeni: I wonder if the branch should have a prefix of `update-` or something

obal/data/roles/fetch_sources/tasks/main.yml
evgeni: Yeah. At some point, we might think about adding a `strategy` switch, defaulting to `upstream`, and behave differently when set to `downstream` (this applies at least to `add` and `update` actions).

But for now the implicit solution is good enough for me :)

evgeni:
@ekohl I think your review is stale and can be dismissed?

ehelms:
Updated with tests

evgeni:
code LGTM :)

ehelms:
@evgeni I think oyu mentioned wanting docs of some sort to go with this. If thats the case, then I'd just wait until the help docs work is in before taking this further.

obal/data/playbooks/update.yml
```diff
@@ -55,3 +48,27 @@
...
+    - name: "Set spec file path, scl options and releasers"
+      set_fact:
+        spec_file_path: "{{ spec_file.files[0].path }}"
+
+    - name: 'Get updated version from spec'
+      command: "rpmspec --query --queryformat=%{version} --srpm {{ spec_file_path }}"
+      register: updated_version
+
+    - name: 'Create git branch'
+      command: "git checkout -b update-{{ inventory_hostname }}-{{ updated_version.stdout }}"
```
ekohl: I'm not a fan of doing git operations. In foreman-packaging I prefer to use `rpm/` and `deb/` prefixes. If obal becomes capable of doing both it can easily conflict as well. Perhaps we can have a branch prefix variable that's empty by default but set in foreman-packaging's manifest?

obal/data/playbooks/update.yml
```diff
@@ -55,3 +48,27 @@
...
       when:
         - "'://' not in item"
+
+    - name: get spec file
+      find:
+        pattern: "*.spec"
+        path: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
+      register: spec_file
+
+    - name: "Set spec file path, scl options and releasers"
```
ekohl: I only see the file path here

obal/data/playbooks/update.yml
```diff
@@ -55,3 +48,27 @@
       with_items: "{{ new_source_urls.stdout_lines | list }}"
       when:
         - "'://' not in item"
+
+    - name: get spec file
```
ekohl: In other places (above) we use `*.spec` in tools. Should we move this part to the top and use the explicit variable?

evgeni:
@ehelms I am mostly concerned about https://github.com/theforeman/obal#updating-a-package-from-downstream having the wrong syntax now, but maybe we can live with it until the new docs are ready.

obal/data/playbooks/update.yml
ehelms: I could base it off current branch and include that in the branch name.

ekohl: I think I prefer explicit configuration over autodetection in this case for simplicity, but maybe the autodetection isn't that hard to write.

obal/data/playbooks/update.yml
```diff
@@ -55,3 +58,21 @@
       with_items: "{{ new_source_urls.stdout_lines | list }}"
       when:
         - "'://' not in item"
+
+    - name: 'Get updated version from spec'
+      command: "rpmspec --query --queryformat=%{version} --srpm {{ spec_file_path }}"
+      register: updated_version
+
+    - name: 'Get current git branch'
+      command: "git rev-parse --abbrev-ref HEAD"
```
evgeni: this needs a `chdir: "{{ inventory_dir }}"` :)

evgeni:
@ehelms as #126 is merged now, would you mind adding/updating the docs for the action?

ehelms:
Not sure why this is failing the test, I copied the help text directly to the fixture.

tests/fixtures/help/update.txt
```diff
@@ -3,6 +3,13 @@ usage: obal update [-h] [-v] [-e EXTRA_VARS] package [package ...]
 Update a package to a newer version
 
 The newer version is either taken from an upstream repository or can be given via command line options.
+To update a package to a new version from the command line:
```
ekohl: Above this is a newline missing

ehelms: I don't get it though, I literally copied this from the output of `--help` and pasted it in here. That implies to me the testing is not verifying what a user sees?

ehelms: I swear I aint crazy, check this screenshot:

screenshot from 2018-10-10 08-48-07

PR #70, opened 2018-03-08
simplify mockbin logging

evgeni:
* drop the path to the binary, thus eliminating the need to replace `{bin}` everywhere
* log real strings, not lists printed as strings

### DIFF
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -259,8 +259,8 @@ def test_obal_scratch_copr_hello_nowait():
     assert os.path.exists('packages/hello/hello-2.10.tar.gz')
 
     expected_log = [
-        "copr-cli create copr-repo-scratch --chroot epel-7-x86_64 --description Scratch Builds --unlisted-on-hp on --repo http://mirror.centos.org/centos/7/sclo/x86_64/rh/",  # noqa: E501
-        "copr-cli edit-chroot copr-repo-scratch/epel-7-x86_64 --packages scl-utils-build rh-ruby24-build",  # noqa: E501
+        "copr-cli create copr-repo-scratch --chroot epel-7-x86_64 --description 'Scratch Builds' --unlisted-on-hp on --repo http://mirror.centos.org/centos/7/sclo/x86_64/rh/",  # noqa: E501
+        "copr-cli edit-chroot copr-repo-scratch/epel-7-x86_64 --packages 'scl-utils-build rh-ruby24-build'",  # noqa: E501
         "tito build --srpm --scl=copr-scl",
         "copr-cli build --nowait copr-repo-scratch hello.src.rpm"
     ]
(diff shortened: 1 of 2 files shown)


evgeni:
@ekohl while I agree that the old way was more explicit, I found the syntax just too cumbersome to write and read.

tests/test_functional.py
```diff
@@ -260,10 +259,10 @@ def test_obal_scratch_copr_hello_nowait():
     assert os.path.exists('packages/hello/hello-2.10.tar.gz')
 
     expected_log = [
-        "['{bin}/copr-cli', 'create', 'copr-repo-scratch', '--chroot', 'epel-7-x86_64', '--description', 'Scratch Builds', '--unlisted-on-hp', 'on', '--repo', 'http://mirror.centos.org/centos/7/sclo/x86_64/rh/']",  # noqa: E501
-        "['{bin}/copr-cli', 'edit-chroot', 'copr-repo-scratch/epel-7-x86_64', '--packages', 'scl-utils-build rh-ruby24-build']",  # noqa: E501
-        "['{bin}/tito', 'build', '--srpm', '--scl=copr-scl']",
-        "['{bin}/copr-cli', 'build', '--nowait', 'copr-repo-scratch', 'hello.src.rpm']"
+        "copr-cli create copr-repo-scratch --chroot epel-7-x86_64 --description Scratch Builds --unlisted-on-hp on --repo http://mirror.centos.org/centos/7/sclo/x86_64/rh/",  # noqa: E501
```
evgeni: @ehelms fwiw, this line is an example how the user would *not* have run the command, as they would have called

    copr-cli create copr-repo-scratch --chroot epel-7-x86_64 --description 'Scratch Builds' --unlisted-on-hp on --repo http://mirror.centos.org/centos/7/sclo/x86_64/rh/

yet the test framework loses the quotes here

evgeni:
@ehelms @ekohl I think now we have the best of both worlds, whatcha think?

tests/test_functional.py
```diff
@@ -260,10 +259,10 @@ def test_obal_scratch_copr_hello_nowait():
     assert os.path.exists('packages/hello/hello-2.10.tar.gz')
 
     expected_log = [
-        "['{bin}/copr-cli', 'create', 'copr-repo-scratch', '--chroot', 'epel-7-x86_64', '--description', 'Scratch Builds', '--unlisted-on-hp', 'on', '--repo', 'http://mirror.centos.org/centos/7/sclo/x86_64/rh/']",  # noqa: E501
-        "['{bin}/copr-cli', 'edit-chroot', 'copr-repo-scratch/epel-7-x86_64', '--packages', 'scl-utils-build rh-ruby24-build']",  # noqa: E501
-        "['{bin}/tito', 'build', '--srpm', '--scl=copr-scl']",
-        "['{bin}/copr-cli', 'build', '--nowait', 'copr-repo-scratch', 'hello.src.rpm']"
+        "copr-cli create copr-repo-scratch --chroot epel-7-x86_64 --description 'Scratch Builds' --unlisted-on-hp on --repo http://mirror.centos.org/centos/7/sclo/x86_64/rh/",  # noqa: E501
+        "copr-cli edit-chroot copr-repo-scratch/epel-7-x86_64 --packages 'scl-utils-build rh-ruby24-build'",  # noqa: E501
```
ekohl: See, there is a case where it's adding something :)

evgeni: Yeah, never said it's not a valid point ;)

ekohl:
Needs a rebase now due to my ```--test``` change.

PR #75, opened 2018-03-08
Properly set the defaults for scratch_build

ekohl:
The old way didn't allow overriding in package_manifest.yml

### DIFF
--- a/obal/data/roles/build_package/tasks/koji.yml
+++ b/obal/data/roles/build_package/tasks/koji.yml
@@ -12,12 +12,12 @@
 - name: 'Set test build'
   set_fact:
       build_package_tito_args: "{{ build_package_tito_args }} --test"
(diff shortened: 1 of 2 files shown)


evgeni:
Shouldn't defaults go into https://github.com/theforeman/obal/blob/master/obal/data/roles/build_package/defaults/main.yaml?

evgeni:
also this breaks all the tests, as it always does scratch builds now

ekohl:
@evgeni I'd be happy to do that instead.

ekohl:
@evgeni on second thought, we do need different defaults in ```release_package.yml``` than ```scratch_build.yml```. Not sure how to solve this.

ekohl:
Replaced by https://github.com/theforeman/obal/pull/76 & https://github.com/theforeman/obal/pull/77

PR #81, opened 2018-03-14
match action and playbook names

evgeni:


evgeni:
Well, technically speaking they already are, in `data` ;)

@ehelms would you like to see a `data/playbooks` next to `data/roles`? (this would require adjusting `role_path` in `ansible.cfg`, though.

ehelms:
Sure, that extra bit would be nice to make it more obvious but if it complicates things I am fine with them where they live today.

evgeni:
@ekohl like https://github.com/theforeman/obal/issues/82?

ekohl:
@evgeni exactly. I saw it after I commented

PR #85, opened 2018-03-22
Add command to simply bump a release and add changelog

ehelms:


ekohl:
Looks like the tests are not happy

obal/data/playbooks/bump-release.yml
```diff
@@ -0,0 +1,11 @@
+---
+- hosts: packages
+  serial: 1
+  gather_facts: no
+  roles:
+    - package_variables
+  tasks:
+    - name: 'Bump release'
+      command: "rpmdev-bumpspec {{ inventory_hostname }}.spec"
```
evgeni: given we have repos where the package is `tfm-foo`, but the spec is `foo.spec`, should that be a `shell` with `*.spec` instead?

ehelms: It did not like *.spec :/ I'll add spec file detection logic. Stupid f-packaging ;)

evgeni: for `*.spec`, it needs `shell` instead of `command`, but yeah, `find` is good too :+1:

ehelms:
Updated with passing tests, please re-review

ehelms:
This test is not using mockbin for the commands. So it's calling the actual commands. I think there is more value in that than mocking. I could not ascertain an easy way to validate the output that the spec actually changed release.

obal/data/playbooks/bump-release.yml
evgeni: @ehelms pending anwer to the above question about the full path instead of the chdir ;)

ekohl:
You could "grep" for a changelog entry.

evgeni:
You could run `rpmspec -q …` and check the new release number being 2?

bkearney:
I tried this on an internal build, and it did indeed bump the version number. I would have preferred it take in the latest commits in order to create the changelog.

ehelms:
Can you define what you mean by latest commits?

ehelms:
@evgeni @ekohl This was approved in its original form, but I wanted to double check the test updates.

README.md
```diff
@@ -65,6 +65,10 @@ This action installs packages required for proper `obal` usage on your machine.
 
 This action updates a package to a newer version. The newer version is either taken from an upstream repository or can be given via command line options
 
+### bump-release
+
+This action updates a spec file's release field and adds a pre-formatted changelog entry.
```
ekohl: This makes me wonder if we should somehow include this in `obal --help` (but out of scope for this PR).

tests/test_functional.py
```diff
@@ -316,3 +316,10 @@ def test_obal_release_copr_hello():
         "copr-cli watch-build"
     ]
     assert_mockbin_log(expected_log)
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_bump_release_hello():
+    assert_obal_success(['bump-release', 'hello'])
+
+    assert subprocess.check_output(['rpmspec', '-q', '--queryformat=%{release}', '--srpm', 'packages/hello/hello.spec'])  # noqa: E501
```
ekohl: Should we assert a new value? `== '2'` or something?

evgeni: I think I would like that, yeah

ehelms:
@evgeni It passed now!

PR #90, opened 2018-04-30
Only remove sources if remote

ehelms:


obal/data/playbooks/update.yml
```diff
@@ -24,6 +24,8 @@
       args:
         chdir: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
       with_items: "{{ old_source_urls.stdout_lines | list }}"
+      when:
+        - "'://' not in item"
```
ekohl: Isn't this the inverse of what we want? This removes all the local files

ehelms: Solid catch. Updated.

PR #91, opened 2018-05-04
Use Ansible 2.5 tests

ekohl:
https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.5.html#jinja-tests-used-as-filters

### DIFF
--- a/obal/data/playbooks/add.yml
+++ b/obal/data/playbooks/add.yml
@@ -51,4 +51,4 @@
           when: git_status.stdout != ''
           args:
             chdir: "{{ inventory_dir }}"
-      when: package_directory_result|changed
+      when: package_directory_result is changed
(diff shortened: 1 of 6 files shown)


evgeni:
is that breaking compat with ansible 2.4 and older?

ekohl:
I can change `install_requires` to `ansible >= 2.5`. That's probably the best way.

ekohl:
Updated with the `install_requires` bump.

evgeni:
Aye. So I must admit I dislike to force our users to have "the latest" ansible. How about we wait till 2.6 is released and then merge this?

ekohl:
Given obal is not used by a lot of people and 2.5 is in all current RH-based distros I'm leaning to forcing here. I guess we could look at disabling deprecations but that would hide them until it's too late as well.

evgeni:
ok, let's see what that breaks :D

PR #94, opened 2018-05-07
Use regex_search over grepping

ekohl:
This prevents failures in big red letters where actually nothing is
wrong.

Untested for now.


komidore64:
Is this PR somethings that's still desired or is it closeable?

ekohl:
I'd still like something like this to avoid the whole red blob in my logs, but I don't know if I have time to finish it. `ignore_errors` is still evil and should be avoided if possible.

PR #99, opened 2018-05-30
build_package_tito_releaser_args is a join now

pcreech:


obal/data/roles/setup_sources/tasks/git.yml
```diff
@@ -13,4 +13,4 @@
 
 - name: 'Register source dir tito argument'
   set_fact:
-    build_package_tito_releaser_args: '--arg source_dir={{ setup_sources_git_checkout_dir }}'
+    build_package_tito_releaser_args: ['--arg source_dir={{ setup_sources_git_checkout_dir }}']
```
ehelms: I think cleaner syntax would be:

```
build_package_tito_releaser_args:
  - '--arg...
```

PR #103, opened 2018-06-05
Allow downloading RPMs from builds into a repo

ehelms:


ekohl:
Looks like it broke the tests but other than that it looks nice.

ehelms:
Tests passing now

obal/data/roles/build_package/tasks/download.yml
```diff
@@ -1,5 +1,23 @@
...
+        path: "{{ inventory_dir }}/downloaded_rpms"
+
+    - name: "Download {{ build_package_koji_command }} RPM(s)"
+      command: "{{ build_package_koji_command }} download-task {{ item }}"
+      args:
+        chdir: "{{ inventory_dir }}/downloaded_rpms"
+      with_items: "{{ build_package_koji_tasks }}"
+
+    - name: 'Create repo from downloaded RPMs'
+      command: 'createrepo .'
```
ekohl: Is there a difference between running `createrepo .` with chdir vs `createrepo {{ inventory_dir }}/downloaded_rpms`?

obal/data/roles/build_package/tasks/download.yml
```diff
@@ -1,5 +1,23 @@
...
   args:
     chdir: "{{ inventory_dir }}"
-  when: build_package_build_system == 'koji'
+  when: build_package_build_system == 'koji' and build_package_download_logs
+
+- block:
+    - name: 'Create RPM downloads directory'
+      file:
+        state: directory
+        path: "{{ inventory_dir }}/downloaded_rpms"
```
ekohl: Should this be a variable since we use it multiple times?

obal/data/roles/build_package/tasks/download.yml
komidore64: I don't think there is.

PR #105, opened 2018-06-14
check the specfile for a new changelog during lint

komidore64:


obal/data/playbooks/library/rpmspec_changelog_check.py
```diff
@@ -0,0 +1,55 @@
...
+        return result
+
+    specdir = module.params['directory']
+
+    specfile = [ os.path.join(specdir, filename) for filename in os.listdir(specdir) if fnmatch.fnmatch(filename, "*.spec") ][0]
+    changelog_top = subprocess.check_output(['rpm', '--query', '--changelog', '--specfile', specfile]).split("\n")[0]
+    changelog_version_release = changelog_top.split(" ")[-1]
+
+    specfile_version_release = subprocess.check_output(['rpm', '--query', '--queryformat', '%{version}-%{release}\n', '--specfile', specfile]).split("\n")[0]
+    specfile_version_release = specfile_version_release.split(".")[:-1] # trim off the dist
```
ekohl: In other places I explicitly set the dist to empty. I think I used `rpmspec --undefine dist` which removes the need for splitting and joining.

obal/data/playbooks/library/rpmspec_changelog_check.py
```diff
@@ -0,0 +1,55 @@
...
+    specfile_version_release = ".".join(specfile_version_release)
+
+    result['changelog_version_release'] = changelog_version_release
+    result['specfile_version_release'] = specfile_version_release
+
+    if not changelog_version_release == specfile_version_release:
```
ekohl: `if x != y` is probably easier to read

obal/data/playbooks/library/rpmspec_changelog_check.py
```diff
@@ -0,0 +1,55 @@
...
+        argument_spec=module_args,
+        supports_check_mode=True
+    )
+
+    if module.check_mode:
+        return result
+
+    specdir = module.params['directory']
+
+    specfile = [ os.path.join(specdir, filename) for filename in os.listdir(specdir) if fnmatch.fnmatch(filename, "*.spec") ][0]
```
ekohl: Would it be easier to use `glob.glob(os.path.join(specdir, '*.spec'))`? I think that's equivalent.

obal/data/playbooks/library/rpmspec_changelog_check.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/python
```
ekohl: Using `#!/usr/bin/env python` is more common. That way it also works in a virtualenv where the python is in a different location.

obal/data/playbooks/library/rpmspec_changelog_check.py
komidore64: oh my goodness that's so much easier!

obal/data/playbooks/lint.yml
```diff
@@ -9,6 +9,10 @@
       command: "rpmlint {{ inventory_dir }}/{{ package_base_dir }}{{ inventory_hostname }}"
       register: output
 
+    - name: "Check for a new changelog entry"
+      rpmspec_changelog_check:
+        directory: "{{ inventory_dir }}/{{ package_base_dir }}{{ inventory_hostname }}"
```
ehelms: This does not require changing unless you feel it makes sense. A general comment that for something like a `rpmspec_changelog_check` I'd expect to pass it a spec file not a directory containing a spec file since this acts upon a specfile. Or potentially one or the other.

komidore64: I was mostly sticking with the way that `rpmlint` is working immediately above -- taking a directory rather than a path to a specfile.

obal/data/playbooks/library/rpmspec_changelog_check.py
```diff
@@ -0,0 +1,51 @@
...
+    specfile = glob.glob(os.path.join(module.params['directory'], "*.spec"))[0]
+
+    cmd = ['rpm', '--query', '--changelog', '--specfile', specfile]
+    result['changelog_version_release'] = subprocess.check_output(cmd).split("\n")[0].split(" ")[-1]
+
+    cmd = ['rpm', '--query', '--queryformat', '%{version}-%{release}\n', '--undefine', 'dist', '--specfile', specfile]
+    result['specfile_version_release'] = subprocess.check_output(cmd).split("\n")[0]
+
+    if result['changelog_version_release'] != result['specfile_version_release']:
+        module.fail_json(msg="changelog has not been updated for the current version-release", **result)
```
ehelms: I feel that most developers would not fully know what to do with this statement, and that we can potentially lead them in the right direction with something like:

"changelog entry missing for %s" % result['specfile_version_release']

obal/data/playbooks/library/rpmspec_changelog_check.py
```diff
@@ -0,0 +1,51 @@
...
+
+    if module.check_mode:
+        return result
+
+    specfile = glob.glob(os.path.join(module.params['directory'], "*.spec"))[0]
+
+    cmd = ['rpm', '--query', '--changelog', '--specfile', specfile]
+    result['changelog_version_release'] = subprocess.check_output(cmd).split("\n")[0].split(" ")[-1]
+
+    cmd = ['rpm', '--query', '--queryformat', '%{version}-%{release}\n', '--undefine', 'dist', '--specfile', specfile]
```
ekohl: If you leave out the `\n` at the end of the format, can you remove the split on the next line?

komidore64: True, but then i get `0.4.5-20.4.5-20.4.5-20.4.5-2` because I guess `rpm` parses every instance in the spec? I'm not sure why it gets output multiple times.

ehelms: If you add `--srpm` it will not output the version-release for both and give you a single entry.

ekohl: Indeed, otherwise it will print it for every subpackage (like `-doc`) as well.

obal/data/playbooks/library/rpmspec_changelog_check.py
komidore64: updated

ehelms:
Works great! I think addressing the use of `--srpm` in the comments above will clean up the code so what's that is in we can merge this up.

komidore64:
it doesn't look like `rpm` has a `--srpm` flag.

ehelms:
Try switching to `rpmspec` tool

komidore64:
@ehelms @ekohl updated!

obal/data/playbooks/library/rpmspec_changelog_check.py
```diff
@@ -0,0 +1,51 @@
...
+
+    module = AnsibleModule(
+        argument_spec=module_args,
+        supports_check_mode=True
+    )
+
+    if module.check_mode:
+        return result
+
+    specfile = glob.glob(os.path.join(module.params['directory'], "*.spec"))[0]
```
ekohl: You could be very safe and do a `try: except IndexError: module.fail_json(msg='Could not find spec file', **result)` here

obal/data/playbooks/library/rpmspec_changelog_check.py
```diff
@@ -0,0 +1,51 @@
...
+
+    cmd = ['rpm', '--query', '--changelog', '--specfile', specfile]
+    result['changelog_version_release'] = subprocess.check_output(cmd).split("\n")[0].split(" ")[-1]
```
ekohl: You could wrap this up in a `try: except subprocess.CalledProcessError`  (not sure that's the exact exception).

komidore64:
@ekohl @ehelms updated

komidore64:
yeah! tests!

PR #106, opened 2018-06-15
Run createrepo on full path

ehelms:


tests/test_functional.py
```diff
@@ -212,7 +212,7 @@ def test_obal_scratch_downstream_hello_wait_download_rpms():
     expected_log = [
         "tito release obaltest-scratch-rhel-7 -y",
         "brew watch-task",
-        "createrepo ."
+        "createrepo {pwd}/download_rpms"
```
ekohl: This should be `downloaded_rpms` instead

PR #109, opened 2018-07-10
Add changelog create command

ehelms:


komidore64:
Does this break `rpmspec_changelog_check`? It is a module, but still currently lives in the default location of `obal/data/playbooks/library/`.

komidore64:
@ehelms @pcreech bump

PR #110, opened 2018-07-11
Introduce a nightly action

ekohl:


obal/data/playbooks/nightly.yml
```diff
@@ -0,0 +1,8 @@
+---
+- hosts: packages
+  serial: 1
+  gather_facts: no
+  roles:
```
ehelms: Should this be using the nightly role?

obal/data/roles/nightly/tasks/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: "Build a nightly package"
```
ehelms: Feels like a role might be overkill and this could be done in the nightly playbook.

obal/data/playbooks/nightly.yml
ekohl: I guess this might be my copy-paste error and why it didn't work.

obal/data/roles/nightly/tasks/main.yml
ekohl: I wasn't entirely sure how it would work with "hosts" and variables.

ekohl:
Simplified to just a playbook and passing tests

PR #111, opened 2018-07-12
Improve the changelog module

ekohl:
* Follows importing conventions
* Doesn't override the built in file method
* Reports whether it was actually changed
* Optimizes the loop iteration


obal/data/modules/changelog.py
```diff
@@ -33,20 +35,23 @@ def main():
     release = release.strip().replace('"', '')
     version = version.strip().replace('"', '')
 
-    lines = ''
-    with open(spec) as file:
-        lines = file.readlines()
+    with open(spec) as spec_fp:
```
ehelms: what is `fp` ?

ekohl: File Pointer. I often use not sure what else to call it but [file](https://docs.python.org/2.7/library/functions.html#file) is a built in so I'd argue it's worse :)

ehelms: Ideas:

 * spec_file
 * specfile

ekohl: Updated

PR #114, opened 2018-07-17
fix obal setup

komidore64:


ehelms:
You could simplify the logic potentially to add an implicit 'all' if the command coming in is setup I think? Or document it as `obal setup all`.

obal/__init__.py
```diff
@@ -70,15 +69,24 @@ def obal_argument_parser(actions, package_choices):
...
-                        help="""which action to execute""")
 
-    parser.add_argument('package',
-                        metavar='package',
-                        choices=package_choices,
-                        nargs='+',
-                        help="the package to build")
+def obal_argument_parser(actions, package_choices):
+    parser = argparse.ArgumentParser()
+    add_bal_arguments(parser)
```
ehelms: typo causing failing tests

komidore64:
i can't figure out why this is blowing up. if i run `obal` with `epdb` in there it looks like it's dealing with the args correctly. @evgeni HALP!

evgeni:
Reading diffs on a OnePlus3 is challenging...

I *think* the issue is: you made the `action` a subcommand/subparser, but the tests still pass the args to the main command.

The tests (pretend to) call `obal -v scratch foo` but with the subparser you need `obal scratch -v foo`, so that `-v` is an arg for scratch, not obal?

evgeni:
Oh, and instead of doing that for loop to add the same args to all parsers, you might want to use parent parsers.

https://docs.python.org/3/library/argparse.html#parents

komidore64:
what

evgeni:
@komidore64 travis has a weird $HOME and we pass params to "fix" that: https://github.com/theforeman/obal/blob/master/tests/test_functional.py#L19

Probably that needs moving too.

evgeni:
the diff looks good on my phone. will have a closer look tomorrow in the office.

komidore64:
yeeeeeaaaaah! @evgeni thank you for the suggestion!

obal/__init__.py
```diff
@@ -70,15 +69,24 @@ def obal_argument_parser(actions, package_choices):
...
-                        nargs='+',
-                        help="the package to build")
+def obal_argument_parser(actions, package_choices):
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(dest='action',
+                                       help="""which action to execute""")
+    subparsers.required = True
+
+    setup_action_subparser = subparsers.add_parser('setup')
+    add_obal_arguments(setup_action_subparser)
```
evgeni: I kind of disliked this `add_obal_arguments` func, and came up with the following, what do you think?

```
diff --git a/obal/__init__.py b/obal/__init__.py
index 489e873..65090eb 100755
--- a/obal/__init__.py
+++ b/obal/__init__.py
@@ -40,53 +40,50 @@ def find_packages(inventory_path):
     return package_choices
 
 
-def add_obal_arguments(parser):
-    parser.add_argument('-e', '--extra-vars',
+def obal_argument_parser(actions, package_choices):
+    parser = argparse.ArgumentParser()
+
+    parent_parser = argparse.ArgumentParser(add_help=False)
+    parent_parser.add_argument('-e', '--extra-vars',
                         dest="extra_vars",
                         action="append",
                         default=[],
                         help="""set additional variables as key=value or
                         YAML/JSON, if filename prepend with @""")
-    parser.add_argument("-v", "--verbose",
+    parent_parser.add_argument("-v", "--verbose",
                         action="count",
                         dest="verbose",
                         help="verbose output")
-    parser.add_argument("--step",
+    parent_parser.add_argument("--step",
                         action="store_true",
                         dest="step",
                         default=False,
                         help="interactive: confirm each task before running")
-    parser.add_argument('-t', '--tags',
+    parent_parser.add_argument('-t', '--tags',
                         dest='tags',
                         default=[],
                         action='append',
                         help="""only run plays and tasks tagged with these
                         values""")
-    parser.add_argument('--skip-tags',
+    parent_parser.add_argument('--skip-tags',
                         dest='skip_tags',
                         default=[],
                         action='append',
                         help="""only run plays and tasks whose tags do not
                         match these values""")
 
-
-def obal_argument_parser(actions, package_choices):
-    parser = argparse.ArgumentParser()
     subparsers = parser.add_subparsers(dest='action',
                                        help="""which action to execute""")
     subparsers.required = True
 
-    setup_action_subparser = subparsers.add_parser('setup')
-    add_obal_arguments(setup_action_subparser)
-
-    for action in [action for action in list(actions) if action != 'setup']:
-        action_subparser = subparsers.add_parser(action)
-        action_subparser.add_argument('package',
-                                      metavar='package',
-                                      choices=package_choices,
-                                      nargs='+',
-                                      help="the package to build")
-        add_obal_arguments(action_subparser)
+    for action in actions:
+        action_subparser = subparsers.add_parser(action, parents=[parent_parser])
+        if action != 'setup':
+            action_subparser.add_argument('package',
+                                          metavar='package',
+                                          choices=package_choices,
+                                          nargs='+',
+                                          help="the package to build")
 
     if argcomplete:
         argcomplete.autocomplete(parser)
```

komidore64:
@evgeni updated based on your comments.

PR #117, opened 2018-07-24
build srpms locally with `obal srpm <package_name>`

komidore64:


obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -0,0 +1,30 @@
+---
+- name: 'Parse tito.props'
+  tito_props_find_package:
+    package: "{{ inventory_hostname }}"
+    tito_props_path: "{{ tito_props_path }}"
+  register: tito_props
```
ehelms: What about if I am either not using tito or I am not configuring tito in this way?

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -0,0 +1,30 @@
...
+    build_srpm_tito_dist: "--dist {{ tito_props.disttag }}"
+  when: tito_props.disttag is defined
+
+- name: "Define SRPM scl option"
+  set_fact:
+    build_srpm_tito_scl: "--scl {{ tito_props.scl }}"
+  when: tito_props.scl is defined
+
+- name: 'Build SRPM'
+  command: "tito build --srpm --offline {{ build_srpm_tito_dist }} {{ build_srpm_tito_scl }}"
```
ehelms: What situation leads to use needing to do more than simply call this line? Locally on my box I tend to just call this line and everything works without having to specify anything.

obal/data/roles/build_srpm/tasks/main.yaml
komidore64: does obal support build environments where tito isn't used?

obal/data/roles/build_srpm/tasks/main.yaml
komidore64: For instance, I almost always have to specify the `dist` otherwise it builds for `.fcXX` rather than `.elX`.

obal/data/roles/build_srpm/tasks/main.yaml
ehelms: Fair point to the first question, but what about the second?

obal/data/roles/build_srpm/tasks/main.yaml
ehelms: Fair, thought that doesn't seem to hurt anything given the final RPM built will have the correct dist? What about multi-dist targeted packages like katello-host-tools that can be built on el5/6/7/fedora27 ?

komidore64: Ah okay! That's one question I had. "Are there any multi-dist packages?" So I see that the answer is **yes**!

obal/data/roles/build_srpm/tasks/main.yaml
ehelms: I ask because whitelist is specific to koji only (copr and brew releasers don't use it I think).

komidore64: tito cannot operate without a props file

```
$ cd foreman-packaging
$ rm -rf rel-eng
$ cd packages/katello/rubygem-hammer_cli_katello/
$ tito build --srpm --offline --dist .el7 --scm tfm
Creating output directory: /tmp/tito
ERROR: Unable to locate branch configuration: /home/adprice/workspace/foreman-packaging/rel-eng/tito.props
Please run 'tito init'
```

ehelms: Yea, its not the props file itself, sorry for the confusion. It's the contents of the props file. Some of what is in the pops file is releaser specific. I believe the part you are relying on here is Koji releaser specific.

komidore64: Looking at `man tito.props`, I'm seeing that `whitelist` (as well as `disttag`) are used by copr and koji (which would include brew, correct?).

ehelms: BrewReleaser is different than KojiReleaser because of dist-git. You can specify disttag per tag in koji which is independent. But whitelist is not supported by Brew unless this has changed in recent tito. I am still curious what necessitates the complexity over just running ""tito build --srpm"" ?

komidore64: > I am still curious what necessitates the complexity over just running ""tito build --srpm"" ?

I guess I just assumed those flags were necessary, otherwise why do they exist? But If they aren't needed then that can greatly  reduce the complexity of this task.

ehelms: The `--offline` tag was needed previously when using tito tagging to manage packages. It was a way to tell tito "do not reach out to git and confirm this tag exists" for pre-release testing. For scratch builds in Koji, I believe it does not matter if you set --scl properly on the SRPM it only matters when doing a release to koji.

komidore64:
@ehelms greatly simplified!

ehelms:
Simplification is much appreciated and it works great. Mind adding a functional test for it?

komidore64:
@ehelms test added

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- name: 'Build SRPM'
+  command: "tito build --srpm --offline"
```
ekohl: It would be nice if we had a way to build SRPMs for nightly packages as well.

komidore64: It does sound nice, but that seems out of scope for this pull request.

komidore64:
bump

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -0,0 +1,14 @@
...
+  args:
+    chdir: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
+  register: srpm_build
+
+- debug:
+    msg: "{{ srpm_build.stdout_lines }}"
```
evgeni: I'd use `stdout` instead of `stdout_lines` here?

```
stdout

Creating output directory: /tmp/tito
Building package [1.2-3]

Wrote: /tmp/tito/bleh-1.2-3.fc28.src.rpm
```

```
stdout_lines
[u'Creating output directory: /tmp/tito', u'Building package [1.2-3]', u'', u'Wrote: /tmp/tito/bleh-1.2-3.fc28.src.rpm']
```

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -0,0 +1,14 @@
...
+  when: srpm_build is successful
+
+- debug:
```
evgeni: This is never executed in the failure case, as ansible will abort after the `command` step failed.

komidore64:
@evgeni updated!

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- name: 'Build SRPM'
+  command: "tito build --srpm --offline"
+  args:
+    chdir: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
+  register: srpm_build
+
+- debug:
```
evgeni: this is still not reached when `command` fails, as it (the `command` module) will directly bail out. but I think we can actually drop that `debug` alltogether, as `command` will output the error cleanly if there was one.

komidore64:
@evgeni error debug task has been removed

komidore64:
thanks yall!

PR #124, opened 2018-08-21
[RFC] Add help texts to playbooks

ekohl:
This introduces an obal metadata file that describes a playbook.  Currently it can store a help text and a hash of variables. Those will get translated into command line arguments with a help text.

Do we want to to expose more parameters of add_argument?

There are some subtleties in doing that, like not copying defaults. Currently they are only passed in if they are non-empty which means you can't override defaults with an empty value. They are also prefixed with --arg so they don't conflict with the other options but that may be overkill. It does make handling very easy.

This PR now has a metadata example for changelog. This will be rendered as follows:

```
$ obal --help
usage: obal [-h]
            {bump-release,changelog,scratch,setup,lint,update,nightly,repoclosure,add,source,release,cleanup-copr,srpm,check}
            ...

positional arguments:
  {bump-release,changelog,scratch,setup,lint,update,nightly,repoclosure,add,source,release,cleanup-copr,srpm,check}
                        which action to execute
    bump-release
    changelog           The changelog command writes a RPM changelog entry for
                        the current version and release.
    scratch
    setup
    lint
    update
    nightly
    repoclosure
    add
    source
    release
    cleanup-copr
    srpm
    check

optional arguments:
```

```
$ obal changelog --help
usage: obal changelog [-h] [-e EXTRA_VARS] [-v] [--step] [-t TAGS]
                      [--skip-tags SKIP_TAGS] [--arg-changelog ARG_CHANGELOG]
                      package [package ...]

positional arguments:

optional arguments:
  -e EXTRA_VARS, --extra-vars EXTRA_VARS
                        set additional variables as key=value or YAML/JSON, if
                        filename prepend with @
  --step                interactive: confirm each task before running
  -t TAGS, --tags TAGS  only run plays and tasks tagged with these values
  --skip-tags SKIP_TAGS
                        only run plays and tasks whose tags do not match these
                        values
  --arg-changelog ARG_CHANGELOG
                        The text for the changelog entry
```

### DIFF
--- a/obal/__init__.py
+++ b/obal/__init__.py
@@ -6,11 +6,11 @@ from __future__ import print_function
 import argparse
 import errno
 import glob
+import json
 import os
 import sys
 
 import yaml
-
 from pkg_resources import resource_filename
 
 try:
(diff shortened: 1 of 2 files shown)


ekohl:
Actual run output:

```
$ obal changelog tfm --arg-changelog bla

PLAY [packages] **************************************************************************

TASK [get spec file] *********************************************************************
Tuesday 21 August 2018  19:34:07 +0200 (0:00:00.097)       0:00:00.097 ******** 
ok: [tfm]

TASK [ensure changelog] ******************************************************************
Tuesday 21 August 2018  19:34:07 +0200 (0:00:00.287)       0:00:00.385 ******** 
changed: [tfm]

PLAY RECAP *******************************************************************************
tfm                        : ok=2    changed=1    unreachable=0    failed=0   

Tuesday 21 August 2018  19:34:07 +0200 (0:00:00.423)       0:00:00.808 ******** 
=============================================================================== 
ensure changelog ------------------------------------------------------------------ 0.42s
get spec file --------------------------------------------------------------------- 0.29s
(obal) ekohl@wisse ~/dev/foreman-packaging (rpm/develop) $ git diff
diff --git a/packages/foreman/tfm/tfm.spec b/packages/foreman/tfm/tfm.spec
index 45f36f580..08a7f33f1 100644
--- a/packages/foreman/tfm/tfm.spec
+++ b/packages/foreman/tfm/tfm.spec
@@ -299,6 +299,9 @@ selinuxenabled && load_policy || :
 %{_root_sysconfdir}/rpm/macros.%{scl_name}-scldevel
 
 %changelog
+* Tue Aug 21 2018 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> 4.0-4
+bla
+
 * Tue Aug 21 2018 Ewoud Kohl van Wijngaarden <ewoud@kohlvanwijngaarden.nl> - 4.0-4
 - Remove rubygem-jquery_pwstrength_bootstrap_4
 
```

obal/__init__.py
```diff
@@ -103,6 +127,9 @@ def generate_ansible_args(inventory_path, playbook_path, args):
         ansible_args.append("-%s" % str("v" * args.verbose))
     for extra_var in args.extra_vars:
         ansible_args.extend(["-e", extra_var])
+    for key, value in vars(args).items():
```
evgeni: would it be easier/nicer to use a custom [Action](https://docs.python.org/3/library/argparse.html#argparse.Action) to store this?

obal/__init__.py
```diff
@@ -22,7 +25,23 @@ def find_playbooks(playbooks_path):
...
         filename = os.path.basename(playbook)
         action = os.path.splitext(filename)[0]
-        playbooks_actions[action] = playbook
+
+        metadata = {
+            'help': '',
+            'vars': {},
+        }
+        try:
+            with open(os.path.join(playbooks_path, '{}.obal.yaml'.format(action))) as obal_metadata:
```
evgeni: I'd actually call that file `{playbook}.metadata.yml`.

also: cheater for using `.yaml` to trick the `glob` above :P

evgeni:
Related:
* https://github.com/ansible/proposals/issues/19
* https://clouddocs.f5.com/products/orchestration/ansible/devel/development/playbook-metadata.html

pcreech:
So, just to capture here, so I don't lose it:

i'm not a fan of adding another 'yaml' extension in the playbook directory, as this will inevitably cause an implicit suggestion that these are also valid playbooks.  Perhaps we should think about options.

obal/__init__.py
```diff
@@ -16,14 +19,69 @@
...
+        self.path = path
+
+        filename = os.path.basename(path)
+        self.name = os.path.splitext(filename)[0]
+
+        self._metadata = None
+
+    @property
+    def _metadata_path(self):
+        return os.path.join(os.path.dirname(self.path), '{}.obal.yaml'.format(self.name))
```
evgeni: thinking aloud: `{}.meta`, `{}.obal`, smth else that is not `.y(a)?ml`?

evgeni:
for comparison, this is how it could look like in plain python:

https://github.com/theforeman/obal/compare/master...evgeni:argparse-cli-new

ekohl:
Closing in favor of https://github.com/theforeman/obal/pull/126

PR #125, opened 2018-08-24
build nightly packages without requiring jenkins artifacts

komidore64:


obal/data/playbooks/nightly.yml
```diff
@@ -2,9 +2,38 @@
...
+  tasks:
+    - set_fact:
+        nightly_noclean: false
+      when: nightly_noclean is not defined
+
+    - name: copy source file
+      copy:
+        src: "{{ item }}"
+        dest: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}/"
+      with_items: "{{ nightly_sourcefiles.split(',') }}"
```
evgeni: `nightly_sourcefiles` is then to be defined on the CLI? tests are failing as that's currently undefined and I don't completely see how this functionality is to be used.

komidore64:
With this latest commit, `obal nightly` will work with both the old-style (passing in `--arg jenkins_job=blah` to `build_package_tito_releaser_args`) as well as the new style outlined in the [package+ breakdown & schedule](https://hackmd.io/35nhQGr6TKmi5ImU7-VM9Q?view).

Will gladly squash after review.

obal/data/playbooks/nightly.yml
```diff
@@ -3,9 +3,28 @@
...
-    releasers:
-      - "{{ nightly_releaser }}"
-    build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
+    build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args | default([]) }}"
+  tasks:
+    - name: 'set nightly_releaser'
+      set_fact:
+        releasers:
+          - "{{ nightly_releaser }}"
+      when: nightly_releaser is defined and "--arg jenkins_job=" in (build_package_tito_releaser_args | join(" "))
```
ehelms: You shouldn't need to do a join here as python should perform an `in` check on an array.

obal/data/playbooks/nightly.yml
```diff
@@ -3,9 +3,28 @@
...
+      - name: 'set build_package_tito_releaser_args'
+        set_fact:
+          build_package_tito_releaser_args:
+            - "{{ build_package_tito_releaser_args }} --arg rpmbuild_options=\"--define 'prerelease .{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}git{{ nightly_githash[:7] }}'\""
+
+      - name: copy source files
+        copy:
+          src: "{{ item }}"
+          dest: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}/"
+        with_items: "{{ nightly_sourcefiles.split(',') }}"
```
ehelms: Could `nightly_sourcefiles` be an array ? And are the sourcefiles expected to be full paths or relative?

obal/data/playbooks/nightly.yml
ekohl: It does make sense because now you can do a sort of fuzzy matching. Otherwise you need an exact match which is impossible to know with a job ID.

komidore64: > You shouldn't need to do a join here as python should perform an `in` check on an array.

It's been a week or so since I wrote this, but I'm fairly certain I tried it without `join(" ")` first. Just to double-check, I'll try it again though.

komidore64: > It does make sense because now you can do a sort of fuzzy matching. Otherwise you need an exact match which is impossible to know with a job ID.

Not sure I'm following you. Mind adding more context?

ekohl: Like I said: I think the current code is the easiest way to achieve what you want so I wouldn't bother :)

ekohl: If it's an array `['--arg jenkins_job=1234']` then `'--arg jenkins_job' in x` will fail while it would work with your joined string.

obal/data/playbooks/nightly.yml
komidore64: You know, I only tested it with absolute paths. I'm not sure about relative.

komidore64: it wants paths to be relative to the playbook which is too complicated to worry about. i'm going to say *no, it doesn't support relative paths* haha

komidore64:
updated!

obal/data/playbooks/nightly/nightly.yaml
```diff
@@ -3,9 +3,32 @@
...
   serial: 1
   gather_facts: no
-  roles:
-    - build_package
   vars:
-    releasers:
-      - "{{ nightly_releaser }}"
-    build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
+    build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args | default([]) }}"
+  tasks:
```
ehelms: If you change this to `pre_tasks` then you can avoid the `import_role` below and just have a `roles` section, e.g.

```
pre_tasks:
 - stufffff
roles:
 - build_package
```

komidore64: oh interesting! is that preferred over `import_role`?

komidore64:
updated again!

PR #126, opened 2018-08-28
Add help texts to playbooks

ekohl:
Includes:
* https://github.com/theforeman/obal/pull/127
* https://github.com/theforeman/obal/pull/128
* https://github.com/theforeman/obal/pull/129


tests/test_obal.py
```diff
@@ -39,21 +43,81 @@ def test_generate_ansible_args_none():
...
+    optional arguments:
+      -h, --help  show this help message and exit
+    """)
+
+    assert help_text == parser.format_help()
+
+
+def test_obal_argument_parser_dummy_help(parser):
+    help_text = textwrap.dedent("""\
+    "usage: pytest dummy [-h] [-v] [-e EXTRA_VARS] [--step] [-t TAGS]
```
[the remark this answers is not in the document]
ekohl: Quite possible yes. I also struggled with this one a bit because it's not really a unit tests but more a functional or at least integration test.

ekohl:
Interesting thought. I was thinking along the lines of:
```
data
├── metadata
│   └── changelog.yaml
└── playbooks
    └── changelog.yaml
```

I don't like this because it's not really obvious they're connected and thus easy to forget.

Another was:

```
data
└── playbooks
    ├── changelog.yaml
    └── metadata
        └── changelog.yaml
```

I think this suffers less from that problem, but still not great.

Your solution makes me think of a third:
```
data
└── playbooks
    └── changelog
        ├── obal.yaml
        └── playbook.yaml
```

Right now I don't know what's the best and I think we could use some more opinions on this.

ekohl:
It's finally green! I believe this is ready for review. Have a look at the help texts and see how it feels.

obal/data/playbooks/release.obal.yaml
```diff
@@ -0,0 +1,5 @@
+help: >
+  This action releases a package to Koji/Brew/Copr
+
+
+  When unless there is already a release the exact same version present no action is performed.
```
komidore64:
```diff
-When unless there is already a release the exact same version present no action is performed.
+No action is performed if a release of the exact same version already exists.
```

ekohl:
> It looks like the package argument is assumed and not included in the help docs. Is that correct?

Package is a positional argument and required.

obal/data/playbooks/release.obal.yaml
ekohl: Agreed

tests/test_playbooks.py
```diff
@@ -21,3 +20,24 @@ def playbook(request):
...
+        obal.main([playbook.name, '--help'])
+    assert excinfo.value.code == 0
+
+    captured = capsys.readouterr()
+
+    help_file = help_dir / '{}.txt'.format(playbook.name)
+
+    try:
+        assert help_file.read() == captured.out
+    except IOError:
```
evgeni: I think this warrants a developer doc, how to record new help files? (I am reading this as: if the file does not exist, it will be written)

evgeni:
I think I like `playbook.yaml` + `obal.yaml` (or `metadata.yaml`?) in the same `action` folder the best.

ekohl:
Pushed additional commits. Can squash them into existing commits later but this might be easier to review.

komidore64:
> > It looks like the package argument is assumed and not included in the help docs. Is that correct?

> Package is a positional argument and required.

Sorry, let me reword.

I am confused by the `package` argument being excluded from the help yaml in this PR. If a command requires it, I would expect it to be listed as a positional argument in the playbook's corresponding help yaml.

The following is displayed as output in the tests:

```
[...]
positional arguments:
[...]
```

That should be linked to some kind of positional argument directive in the help yaml.

Assuming that a command will have a `package` argument and nothing else seems extremely limiting to those who wish to add commands to obal that only want to write ansible. One would have to go add logic into the python for removing the `package` argument to a new command or if they want different arguments.

ekohl:
Initially I was thinking of doing it inside the YAML but metadata is a workaround of too little information in the playbook. With https://github.com/theforeman/obal/pull/131 I changed the hardcoded to some basic auto detection whether it accepts packages or not.

Perhaps we can further improve the introspection to do host mapping but right now we have 1 other case (repoclosures). Trying to come up with a solution now feels like over engineering to me because we don't exactly know the need just yet.

Do you think that is a blocker for this or can we iterate on it?

komidore64:
I'm not a fan of auto-detecting what arguments a playbook may or may not want.

It seems demonstrably simpler to me to have a section in the help yaml for positional arguments describing them and what they should map to:

```
help: >
  The changelog command writes a RPM changelog entry for the current version and release.
variables:
  changelog:
    parameter: --message
    help: The text for the changelog entry
arguments:
  - name: package
    description: the package whose spec will be edited
    variable: inventory_hostname
```

Or something along those lines?

obal/__init__.py
```diff
@@ -47,19 +117,75 @@ def takes_package_parameter(self):
...
+        This is extracted from the metadata.
+        """
+        return self.metadata['variables']
+
+    @property
+    def help_text(self):
+        """
+        The help text if available. This is the first line from the help in the metadata.
+        """
+        return self.metadata['help'].split('\n', 1)[0] if self.metadata['help'] else None
```
evgeni: maybe `.splitlines()` instead of `.split('\n')`?

ekohl:
> you have a few descriptions start with "This action …", which I think should be dropped, as that's IMO implied.

I will admit I mostly copied this from README.md and didn't look at them too carefully so you're probably right.

obal/__init__.py
ekohl: I thought that only splitting it once was more efficient but there's probably little difference

ekohl:
> I'm not a fan of auto-detecting what arguments a playbook may or may not want. Especially when we have plenty of roles involving inheritance.

I do like it when we can reliably do it. That way positional arguments are always a representation of the hosts that are passed to Ansible. You're right that now it's a sort of halfway there solution.

> It seems demonstrably simpler to me to have a section in the help yaml for positional arguments describing them and what they should map to:

Currently this maps into choices which are read from the inventory:
https://github.com/theforeman/obal/blob/9c981309d303a3982e13892c14b7295651cc8abd/obal/__init__.py#L86-L100

This would need to be modified to plug into this.

Given what you propose would new and not changing semantics of the current metadata I'm wondering if we can leave it for a next PR. This one is already getting pretty big.

ekohl:
Added a commit that renames it. Please let me know if this is correct and I'll squash it to a cleaner order.

obal/__init__.py
evgeni: Yeah. I would not think there is much diff. But also would not force you to change :)

ehelms:
I would prefer to add docs only in one place. Auto generating README changes everytime does not sound great either. Using github pages with auto-generated docs would be the best option to me but thats a lot more work. Would it be easy to put instructions on a way to install/clone the repository and run a command that spits out the docs?

ekohl:
Rebased and I believe this can be merged now. When merging this, please don't squash the commits.

ekohl:
In https://github.com/theforeman/obal/pull/135 I did start to add that kind of documentation. Do you want that to be part of this PR?

evgeni:
@ekohl ok, yeah, if there is docs prepared, and just waits on a better docs framework to be merged, I think I can live with it.

ekohl:
Thanks everyone for the review and discussions!

PR #128, opened 2018-08-28
Enable pylint checking and re-enable ansible-lint checking

ekohl:


obal/__init__.py
```diff
@@ -125,7 +125,7 @@ def main(cliargs=None):
     # this needs to be global, as otherwise PlaybookCLI fails
     # to set the verbosity correctly
     from ansible.utils.display import Display
-    global display
+    global display  # pylint: disable=C0103,W0601
```
komidore64: what were these warnings?

ekohl: [C0103](http://pylint-messages.wikidot.com/messages:c0103) is an invalid name. Globals are supposed to be in all caps but this is because ansible requires it. I now see that there are more pylint failures and a better way to clear it. Going to revisit this. Will also look at adding the checks to CI.

ekohl:
Updated. Now it's pylint clean, runs jobs in parallel and runs ansible-lint again. It does look like I broke the coverage. Not sure how.

evgeni:
seems that something broke coverage reporting :/

ekohl:
Fixed the coverage now

evgeni:
@ekohl while coveralls does not complain anymore, it lists 0 files with coverage: https://coveralls.io/jobs/39747198 :/

ekohl:
Re-added the `coverage combine` and that appears to have worked.

PR #129, opened 2018-08-28
Move advanced arguments into their own group

ekohl:


komidore64:
what does the output from this change look like?

ekohl:
```
$ obal add --help
usage: obal add [-h] [-v] [-e EXTRA_VARS] [--step] [-t TAGS]
                [--skip-tags SKIP_TAGS]
                package [package ...]

positional arguments:

optional arguments:

advanced arguments:
  -e EXTRA_VARS, --extra-vars EXTRA_VARS
                        set additional variables as key=value or YAML/JSON, if
                        filename prepend with @
  --step                interactive: confirm each task before running
  -t TAGS, --tags TAGS  only run plays and tasks tagged with these values
  --skip-tags SKIP_TAGS
                        only run plays and tasks whose tags do not match these
                        values
```

evgeni:
I like the new output.

But this made me think: do we still need `--step`, `--tags` and `--skip-tags`? These were exposed because the old playbooks used tags for some steps (wait/download). We don't do that anymore since 1f7dc103c2490cd5e84610e381737e4a2ba96dae. And I'd say the less ansible-specifics in our CLI, the better.

And if we remove them, is it worth keeping `--extra-vars` the only advanced argument? I'd say yes, but wanted to ask the question nevertheless.

ekohl:
> But this made me think: do we still need --step, --tags and --skip-tags? These were exposed because the old playbooks used tags for some steps (wait/download). We don't do that anymore since 1f7dc10. And I'd say the less ansible-specifics in our CLI, the better.

I'd be fine with dropping since I never used these but no strong opinion.

> And if we remove them, is it worth keeping --extra-vars the only advanced argument? I'd say yes, but wanted to ask the question nevertheless.

Yes, even if we only have one, I like the separation. Especially if we start adding more arguments for normal playbooks.

PR #133, opened 2018-09-05
Don't treat existing build as a failure

ehelms:


obal/data/roles/build_package/tasks/wait.yml
```diff
@@ -20,7 +20,7 @@
     - name: 'Failed build'
       fail:
         msg: "The build in {{ build_package_koji_command }}  has failed"
-      when: (build_package_koji_status is failed)
+      when: (build_package_koji_status is failed) and ('Build already exists' not in build_package_koji_status.stdout)
```
ekohl: When it already exists, should we verify it has the right tag?

ehelms: Fair. Will take a bit more logic but I can look into that.

ehelms: I think this is gonna be hard to do @ekohl in our current state. For one, we rely on tito releasers which can mask tags. We have a few ways we denote them but I think we'd need to fundamentally make tags a first class in our packaging variables. The second is around SCL and nonSCL packages. We don't have an easy way to differentiate when we have a dual-citizen. And to look up koji information we need to use the actual build or package name information (e.g. tfm-rubygem-mysql2 and rubygem-mysql) but knowing that is hard given Koji today figures it out for us. There is a small subset that meet this criteria but they do provide edge cases.

evgeni:
I can't remember being anywhere, but LGTM

PR #136, opened 2018-09-18
Remove downstream packages from setup

ekohl:
These only exist in downstream channels. That means running the setup only works in certain environments.


ekohl:
Note that the alternative is to package obal as an RPM and add them as requires so you don't need to run setup. It's what I do with https://copr.fedorainfracloud.org/coprs/ekohl/obal

PR #137, opened 2018-10-08
let the parser be buildable w/o arguments

evgeni:


obal/__init__.py
```diff
@@ -230,10 +230,16 @@ def find_packages(inventory_path):
     return package_choices
 
 
-def obal_argument_parser(playbooks, package_choices):
+def obal_argument_parser(playbooks=None, package_choices=None):
     """
     Construct an argument parser with the given actions and package choices.
     """
+    if not playbooks:
```
ekohl: Perhaps this should be `if playbooks is None` to avoid calling this when an empty list is passed in (and the same below).

evgeni: done

PR #139, opened 2018-10-09
drop --step, --tags and --skip-tags args, we don't use them

evgeni:


evgeni:
well, I guess coverage decrease is expected.

PR #141, opened 2018-10-10
handle RPM specs with epoch

komidore64:


obal/data/modules/changelog.py
```diff
@@ -24,6 +24,18 @@ def en_locale():
             locale.resetlocale(locale.LC_TIME)
 
 
+def format_evr(epoch, version, release):
+    epoch = epoch.strip().replace('"', '')
+    version = version.strip().replace('"', '')
+    release = release.strip().replace('"', '')
+    evr = ""
+
+    if not '(none)' in epoch:
```
ekohl: The common order is `'(none)' not in epoch`.

komidore64: ah okay, thank you

komidore64:
> There's an IMHO cleaner way that rpmdev-bumpspec applies:
> 
> ```python
>         # Get EVR for changelog entry.
>         cmd = ("rpm", "-q", "--specfile", "--define", "dist %{nil}",
>                "--qf=%|epoch?{%{epoch}:}:{}|%{version}-%{release}\n", aspec)
> ```

Oh that's some funky syntax! I looked in the man pages for queryformat help because it didn't like `%{?epoch}` but wasn't able to find anything. I guess it's my fault for not going to read the source ;P

That does look cleaner though.

komidore64:
@ehelms @ekohl updated

ehelms:
Looks good pending test fixes

obal/data/modules/rpmspec_changelog_check.py
```diff
@@ -7,23 +7,49 @@
...
+        'rpm',
+        '--query',
+        '--changelog',
+        '--specfile',
+        specfile
+    ])
+    return evr.split("\n")[0].split(" ")[-1]
+
+def get_specfile_evr(specfile):
+    return subprocess.check_output([
```
evgeni: does this require a `.strip()` at the end to get rid of the `\n` in the output?

obal/data/modules/rpmspec_changelog_check.py
```diff
@@ -7,23 +7,49 @@
...
+    evr = subprocess.check_output([
+        'rpm',
+        '--query',
+        '--changelog',
+        '--specfile',
+        specfile
+    ])
+    return evr.split("\n")[0].split(" ")[-1]
+
+def get_specfile_evr(specfile):
```
evgeni:
```python
>>> get_specfile_evr('/home/remote/egolov/Devel/theforeman/foreman-packaging/packages/foreman/foreman/foreman.spec')
'1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop1.20.0-0.10.develop'
```

this looks wrong :)

evgeni: mhh, pure python:

```
>>> import rpm
>>> rpm.delMacro('dist')
>>> s = rpm.spec('/home/remote/egolov/Devel/theforeman/foreman-packaging/packages/foreman/foreman-installer/foreman-installer.spec')
>>> s.sourceHeader['evr']
'1:1.20.0-0.1.develop'
>>> s = rpm.spec('/home/remote/egolov/Devel/theforeman/foreman-packaging/packages/foreman/foreman/foreman.spec')
>>> s.sourceHeader['evr']
'1.20.0-0.10.develop'
```

be aware, that on python3 this returns `bytes`, not `str`, so it's safest to use 

```
>>> import rpm
>>> rpm.delMacro('dist')
>>> s = rpm.spec('/home/remote/egolov/Devel/theforeman/foreman-packaging/packages/foreman/foreman-installer/foreman-installer.spec')
>>> s.sourceHeader['evr']
b'1:1.20.0-0.1.develop'
>>> s.sourceHeader['evr'].decode('ascii')
'1:1.20.0-0.1.develop'
```

(that should work on python2 too)

obal/data/modules/rpmspec_changelog_check.py
```diff
@@ -7,23 +7,49 @@
...
 ANSIBLE_METADATA = {
-    'metadata_version': '1.1',
+    'metadata_version': '1.2',
     'status': ['preview'],
     'supported_by': 'community'
 }
 
 from ansible.module_utils.basic import AnsibleModule  # pylint: disable=C0413
 
+def get_changelog_evr(specfile):
```
evgeni: pure python version (with the setup like below):

```python
>>> s.sourceHeader[rpm.RPMTAG_CHANGELOGNAME][0].split(" ")[-1]
'1:1.20.0-0.1.develop'
```

obal/data/modules/rpmspec_changelog_check.py
```diff
@@ -40,17 +66,15 @@ def run_module():
...
-        result['specfile_version_release'] = subprocess.check_output(cmd)
+        result['specfile']['epoch_version_release'] = get_specfile_evr(specfile)
     except subprocess.CalledProcessError as err:
         msg = "An error occured while running [ {} ]".format(err.cmd)
         module.fail_json(msg=msg, **result)
 
-    if result['changelog_version_release'] != result['specfile_version_release']:
-        msg = "changelog entry missing for {}".format(result['specfile_version_release'])
+    if result['changelog']['epoch_version_release'] != result['specfile']['epoch_version_release']:
+        msg = "changelog entry missing for {}".format(evr)
```
evgeni: this `evr` seems undefined?

obal/data/modules/rpmspec_changelog_check.py
ekohl: The question is if you want to rely on rpm as a library. It can have odd python 2 & python 3 implications. Depending on it via standard tooling doesn't work because RPM isn't a real python package.

obal/data/modules/rpmspec_changelog_check.py
ekohl: I think it doesn't because the query format doesn't include a newline

ekohl:
In `tests/test_functional.py` we have tests for this. Perhaps we can add a package with epoch to our test suite?

obal/data/modules/rpmspec_changelog_check.py
ehelms: That's a fancy use of library code

obal/data/modules/rpmspec_changelog_check.py
komidore64: Not needed.

```
>>> subprocess.check_output([
... 'rpm',
... '--query',
... '--queryformat',
... '%|epoch?{%{epoch}:}:{}|%{version}-%{release}',
... '--undefine',
... 'dist',
... '--specfile',
... 'foreman-installer.spec'
... ])
'1:1.20.0-0.1.develop'
```

evgeni:
I think you'll have to install `python-rpm` on this Ubuntu box to make the tests pass :)

obal/data/modules/rpmspec_changelog_check.py
```diff
@@ -3,27 +3,30 @@
...
 def run_module():
     module_args = dict(
         directory=dict(type='str', required=True),
     )
 
     result = dict(
-        changed=False,
-        changelog_version_release='',
-        specfile_version_release=''
+        changed = False,
```
ekohl: The python convention is to not use spaces in constructors.

komidore64:
@ehelms @evgeni as nice as it would be to use python's built-in `rpm`, it would be considerably difficult to get a working lib of `rpm` for tox to use.

Shelling out to `rpm` and `rpmspec` may be kinda ugly, but it's still very readable and easy to see what's going on. I don't want to spend days figuring out how to make `python-rpm` play nice with tox just so we don't have to spawn a couple `subprocess` calls.

evgeni:
@komidore64 Did you try to set [`sitepackages=true`](https://tox.readthedocs.io/en/latest/config.html#conf-sitepackages) in `tox.ini`? If that's not fixing it, we'll live with `subprocess` I guess.

ekohl:
I don't mind shelling out because it's a clean separation. RPM isn't really available as a pypi package (which you notice with tox). It's probably also only installed for one python version.

When you do shell out, there's `queryformat=%{evr}` which appears to be exactly what you want.

komidore64:
> When you do shell out, there's `queryformat=%{evr}` which appears to be exactly what you want.

Ugh, that is **so much** prettier!

.travis.yml
```diff
@@ -28,3 +28,4 @@ addons:
     packages:
     - git-annex
     - rpm
+    - python-rpm
```
ehelms: Is this still needed given the revert back to using subprocess?

komidore64: oop, yeah needs to be removed.

.pylintrc
```diff
@@ -1,4 +1,4 @@
 [MASTER]
 ignore=tests
 # Disable useless-object-inheritance because we still support py2
-disable=R0205
+disable=R0205,R0801
```
ekohl: I'm not sure if this should be applied globally or just for our modules. @evgeni?

obal/data/modules/rpmspec_changelog_check.py
```diff
@@ -22,8 +44,12 @@ def run_module():
 
     result = dict(
         changed=False,
-        changelog_version_release='',
-        specfile_version_release=''
+        changelog=dict(
```
ekohl: I'm not sure this change is really needed and adds a lot of readability. I think it mostly makes it longer.

.pylintrc
komidore64: I was unable to get the disable to work anywhere else but here.

evgeni: Well, but `pylint` is right, it is duplicated code. So lets unduplicate it :)

https://gist.github.com/evgeni/18e65f302c44a5dd94251d5b6306f71a

komidore64:
updated again

.pylintrc
ehelms: Does that imply this should be dropped now from the changeset?

evgeni: Err, yeah, of course. Forgot to add that to the patch.

komidore64:
rebased to fix `tox.ini` conflict

obal/data/module_utils/obal.py
```diff
@@ -0,0 +1,28 @@
...
+    return evr.split("\n")[0].split(" ")[-1]
+
+def get_specfile_evr(specfile):
+    """get the EVR from the source header of the specfile"""
+    return subprocess.check_output([
+        'rpmspec',
+        '--query',
+        '--queryformat',
+        '%{evr}',
+        '--undefine',
```
evgeni: we should also undefine `foremandist`, @komidore64 want to do this in this PR, or in another?

.pylintrc
```diff
@@ -1,4 +1,4 @@
 [MASTER]
 ignore=docs,tests
 # Disable useless-object-inheritance because we still support py2
-disable=R0205
+disable=R0205,R0801
```
evgeni: damn, seems I forgot to drop this in the end :/

evgeni:
i made you https://github.com/komidore64/obal/pull/2

tox.ini
```diff
@@ -17,7 +17,7 @@ deps =
     pytest-xdist
     coveralls
 commands =
-    check-manifest --ignore tox.ini,tests*,docs*
+    check-manifest --ignore tox.ini,tests*,docs*,*/**/*.pyc,*/**/__pycache__
```
evgeni: I never needed these, but *shrug*

tests/fixtures/testrepo/upstream_with_epoch/repoclosure/el7.conf
```diff
@@ -0,0 +1,33 @@
+[main]
```
evgeni: do we need this file, or is it here because we have it in the other dummy repos too? I'd (personally) drop it, but would not force you to :)

evgeni:
I'm good with the changes, but would prefer an ACK from @ekohl or @ehelms or @pcreech, as I have made changes in this PR

obal/__main__.py
```diff
@@ -1,5 +1,5 @@
 """
 Main module file. This allows you to run python -m obal
 """
-from obal import main
+from obal import main  # pylint:disable=no-name-in-module
```
ekohl: Odd that I didn't have to do this. How did you install your env?

evgeni: I was using `tox -e py27`

ekohl:
Squashed & merged since I needed it and blocking the 1.20.0-RC1 release. Thanks!

tox.ini
komidore64: I needed them after the initial run of `tox`.

komidore64:
@ekohl thank you!

PR #144, opened 2018-10-17
Use version to disable pulling from upstream during an update

ehelms:


evgeni:
The tests fail because the packaging repo stubs we ship are no git repositories.

There is some code that does a git setup, but it's not used universally:
https://github.com/theforeman/obal/blob/cabab363710411341d421102149bd03034df2ab6/tests/test_functional.py#L74-L84

Other than that, I think `version` is a valid differentiator for the "do update" vs "pull update" scenarios.

evgeni:
https://github.com/theforeman/obal/pull/145 should fix the tests for you.

evgeni:
Given no tests broke when you modified this behavior in #68, I think we should add tests for this branch of logic now?

PR #145, opened 2018-10-18
properly init the git we have in the packaging repo

evgeni:
sometimes `git init` is not enough and you need to commit stuff to have
a branch and everything.

### DIFF
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -20,6 +20,13 @@ if os.environ.get('TRAVIS', None):
     DEFAULT_ARGS.extend(['-e', 'ansible_remote_tmp=/tmp/ansible-remote'])
 
 
+def git_init(path):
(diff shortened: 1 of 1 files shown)


tests/test_functional.py
```diff
@@ -20,6 +20,13 @@
     DEFAULT_ARGS.extend(['-e', 'ansible_remote_tmp=/tmp/ansible-remote'])
 
 
+def git_init(path):
+    subprocess.check_call(['git', 'init'], cwd=path)
+    subprocess.check_call(['git', 'annex', 'init'], cwd=path)
+    subprocess.check_call(['git', 'add', '.'], cwd=path)
+    subprocess.check_call(['git', 'commit', '-a', '-m', 'init'], cwd=path)
```
ekohl: The `-a` is redundant after the explicit `git add` but doesn't hurt either.

PR #149, opened 2018-11-01
nightly - add options

komidore64:


obal/data/playbooks/nightly/metadata.obal.yaml
```diff
@@ -3,3 +3,11 @@ help: >
...
 
   This typically uses unreleased sources taken from a git repository.
+variables:
+  nightly_sourcefiles:
+    help: Comma-separated list of absolute filepaths to be copied and used for RPM sources.
+  nightly_githash:
+    help: Git hash to be included in the package's release.
+  build_package_scratch:
+    parameter: --scratch
+    help: Set 'true' to scratch build.
```
ekohl: I feel like we should expand the metadata a bit so we can set this to a boolean type but I'm fine with starting here.

obal/data/playbooks/nightly/metadata.obal.yaml
```diff
@@ -3,3 +3,11 @@ help: >
 
 
   This typically uses unreleased sources taken from a git repository.
+variables:
+  nightly_sourcefiles:
+    help: Comma-separated list of absolute filepaths to be copied and used for RPM sources.
```
ekohl: Is the native format also comma separated? Might be good to investigate moving to an array but that means this also needs to change so :+1: to start with this

komidore64: I don't think there's stuff currently in our help mechanisms for arrays. but you're right, i'd need to change the `nightly` playbook a bit also.

komidore64:
i'm not entirely sure what's failing, though i can reproduce locally.

evgeni:
Proposed fix in https://github.com/theforeman/obal/pull/150

komidore64:
@ekohl rebased!

komidore64:
waiting on #151

evgeni:
#151 was merged, feel free to use that feature now

komidore64:
thannks @evgeni.

this PR has been rebased to include #151.

obal/data/playbooks/nightly/metadata.obal.yaml
komidore64: #151 has been merged, so `--scratch` now has:

```
action: store_true
```

obal/data/playbooks/nightly/metadata.obal.yaml
ekohl: We now have `action: append` which allows you to repeat the argument, in case you want to modify the playbook.

komidore64:
@ekohl i was just looking into that! thanks :)

ekohl:
https://obal.readthedocs.io/en/latest/development.html#exposing-playbooks-using-metadata even has a nice reference example :)

komidore64:
@ekohl updated!

PR #150, opened 2018-11-02
make the params generated from the variables sorted

evgeni:


komidore64:
this fixed the failures i was seeing in #149.

PR #151, opened 2018-11-02
Allow setting the action for parameters

ekohl:
Currently only store_true and store_false are implemented.


komidore64:
I appreciate that this solves my boolean flag problem, but this seems like a lot of code just to facilitate the equivalent of:

```python
parser.add_argument('--foo', action='store_true')
```

ekohl:
I agree it's more code than I'd like, but we need to store all Ansible variables in a special way so we can pass in all variables. I've copied the actions from argparse: https://github.com/python/cpython/blob/e2ed5adcb5db2d70cfa72da1ba8446f7aa9e05cd/Lib/argparse.py#L874-L928

The action map is the equivalent of a registry in argparse: https://github.com/python/cpython/blob/e2ed5adcb5db2d70cfa72da1ba8446f7aa9e05cd/Lib/argparse.py#L1256-L1258

The alternative is to store them in the regular way but then to collect them, but I think that has a risk of being just as much (or more) code. Could also be more fragile.

Maybe @evgeni has an idea on how to reduce the amount of code that's needed, but I don't really see how.

pcreech:
So, my mind goes to why we are explicitly defining classes that are almost 1-1 what is defined in argparse.

It seems it's to explicilty support the following construct:

```
if hasattr(args, 'variables'):
        ansible_args.extend(["-e", json.dumps(args.variables, sort_keys=True)])
```
I don't have an immediate solution, but perhaps there could be a better way to generate this "-e {json}" format.  This way, we wouldn't have to keep re-defining argparse constructs just to expose argparse provided data types.

The immediate simplistic solution that comes to mind is to create a registry of playbook variable names.  We are already parsing every playbook as part of `obal_argument_parser()`.  If we create a kind of registry of `arguments[playbook.name] = playbook.playbook_variables` we could then have either a new function, or a section of `generate_ansible_args`, that would do something like this:

```
playbook_arguments = dict()
for argument in arguments[plabook.name]:
    playbook_arguments[argument] = getattr(args, argument)
ansible_args.extend(['-e', json.dumps(playbook_arguments, sort_keys=True))
```
In general, I'm leaning towards building out a better way of generating '-e' inside `generate_ansible_args()` that will allow us to better take advantage of what is already provided by argparse, to allow us to prevent needing boilerplate redefinition, and to open up argparse's entire 'action' collection to a playbook variable.

I'm not entirely sure if that's truly feasible, but everything in the code so far looks like we should be able to do it.

ekohl:
I've added a commit that stores a list of "obal args" that we later extract and store in a dict. This removes the need for a custom action and allows us to use the built in actions. In a very similar way as @pcreech describes.

Not sure if there's a better place to store this list than on the parser, but I think it's not too ugly.

evgeni:
If we set `dest` to `_obal_%(variable.name)`, we could filter on this prefix when collecting and wouldn't need the extra list.

Not sure that'd be cleaner though.

ekohl:
> Not sure that'd be cleaner though.

Turns out the answer is no, it would not be prettier. dest is also used in the help text so then you need to use a metavar. Metavars are not allowed on actions that do not accept parameters (like `store_true` and `store_false`). I think detecting that is more fragile and uglier than the current approach.

komidore64:
This looks much easier to follow.

ekohl:
I'm also much happier with the reduction of duplication.

ekohl:
Any objects left to merge this? Shall I squash the commits?

evgeni:
I'll squash when merging, thanks!

PR #154, opened 2018-11-13
implement waiting for a package to be published into the repo

evgeni:


obal/data/roles/build_package/defaults/main.yaml
```diff
@@ -8,3 +8,4 @@ build_package_tito_releaser_args: []
 build_package_wait: true
 build_package_download_logs: false
 build_package_download_rpms: false
+build_package_waitrepo: false
```
ehelms: I would honestly love if this were true by default.

obal/data/roles/build_package/tasks/waitrepo.yml
```diff
@@ -0,0 +1,14 @@
...
+    register: build_package_koji_task_info
+
+  - name: "Filter {{ build_package_build_system }} output"
+    set_fact:
+      build_package_koji_task_data: "{{ build_package_koji_task_info.results | map(attribute='stdout') | map('from_yaml') | list }}"
+
+  - name: "Wait for {{ build_package_koji_command }} to regenerate the repo"
+    command: "{{ build_package_koji_command }} wait-repo --build={{ item['Build'] | regex_replace(' .*', '') }} {{ item['Request Parameters']['Build Target'] | regex_replace('-candidate$', '') }}-build"
+    with_items: "{{ build_package_koji_task_data }}"
+  when: build_package_build_system == 'koji' and build_package_waitrepo|bool
```
ehelms: Can this be placed above the block? e.g.

```
- when: blahl blahb blahb
  block:
    -
```

I feel like given yaml this should work and is in my opinion much more readable than the dangling when

obal/data/roles/build_package/defaults/main.yaml
evgeni: We can do that, the only dependency is build_wait, and that's already true by default.

obal/data/roles/build_package/tasks/waitrepo.yml
evgeni: My memory says it does not, but I'll try out to be sure tomorrow.

evgeni:
@ehelms updated!

obal/data/roles/build_package/tasks/waitrepo.yml
evgeni: It did \o/

obal/data/roles/build_package/tasks/waitrepo.yml
```diff
@@ -0,0 +1,14 @@
+- when: build_package_build_system == 'koji' and build_package_waitrepo|bool
```
ehelms: Nice! We should do this more often, feels more like a traditional `if-then` relationship to me.

PR #155, opened 2018-11-13
obal nightly - branching conditional was incorrect

komidore64:


obal/data/playbooks/nightly/nightly.yaml
```diff
@@ -15,7 +15,7 @@
 
         - import_role:
             name: build_package
-      when: nightly_releaser is defined and '--arg jenkins_job=' in (build_package_tito_releaser_args | join(' '))
+      when: nightly_sourcefiles is not defined and nightly_githash is not defined
```
evgeni: Shouldn't this *still* check for `nightly_releaser` being defined? As otherwise the `set_fact` would bomb out?

evgeni:
I can live with that then :)

evgeni:
Now, looking at it, the `set_fact` is then redundant to the `package_manifest` entry? As that already sets `releasers: nightly_releaser`

evgeni:
(the ack still stands, as the comment only says the set fact is redundant ;))

evgeni:
OK, this test failure was unexpected.

evgeni:
hah. so our tests don't set `releasers` :D

https://github.com/theforeman/obal/blob/1f660d5fc7f4ee35d33945213632297714b3f488/tests/fixtures/testrepo/upstream/package_manifest.yaml#L14-L16

vs

https://github.com/theforeman/foreman-packaging/blob/c8749598adec11f182554543104f12f918cff5c1/package_manifest.yaml#L34-L39

komidore64:
pfffffft

komidore64:
@evgeni updated!

tests/test_functional.py
```diff
@@ -159,10 +160,10 @@ def test_obal_release_upstream_hello_nowait():
 
 @obal_cli_test(repotype='upstream')
 def test_obal_nightly_upstream_hello():
-    assert_obal_success(['nightly', 'hello'])
+    assert_obal_success(['nightly', 'hello', '--source', os.path.join(MOCK_SOURCES_DIR, 'hello-2.10.tar.gz'), '--githash', '0123456789abcdef'])
 
     expected_log = [
-        "tito release dist-git-jenkins -y --arg jenkins_job=hello-master-release",
+        "tito release dist-git -y --arg jenkins_job=hello-master-release",
```
ekohl: If you pass in the explicit source path it surprises me we still pass in the jenkins_job arg. I'd also like to verify in some way we actually pass in the git hash in the version but not sure how.

obal/data/playbooks/nightly/nightly.yaml
```diff
@@ -8,14 +8,14 @@
   tasks:
     - name: 'legacy nightly building'
       block:
-        - name: 'set nightly_releaser'
```
ekohl: Why break this? Modern building shouldn't use this.

tests/test_functional.py
komidore64: the jenkins_job arg is sitll "passed" in because the variable still exists in the package_manifest. the `build_package` role still populates the variable, but only the `-jenkins` releasers now how to deal with that arg, so it's a no-op.

obal/data/playbooks/nightly/nightly.yaml
komidore64: I am not following you.

ekohl: Is there a reason you're removing this code? As I understand it the new way of building shouldn't hit this code so it should be safe to keep it in.

ekohl: Another thought: perhaps move it into a role and write all the legacy building in one file while doing the modern way in another file. IMHO it's easier to follow than huge blocks with when statements.

komidore64: the new way of building shouldn't hit this, correct. but it's redundant to have for legacy nightly building.

RE: roles
that's also a good idea and i agree with you, but sounds more like refactoring work.

komidore64: This task has been returned.

PR #162, opened 2018-12-13
add --dist and --scl to `obal srpm`

komidore64:


obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -1,10 +1,41 @@
 ---
+- name: 'Set build_srpm_tito_releaser_args'
+  set_fact:
+    build_srpm_tito_releaser_args: "{{ build_package_tito_releaser_args | default([]) }}"
```
evgeni: you define `build_srpm_tito_releaser_args: []` in `defaults`, which makes me think the user is allowed to set `build_srpm_tito_releaser_args: ['--whatever']`. but this `set_fact` will override it?

obal/data/roles/build_srpm/defaults/main.yaml
```diff
@@ -0,0 +1,5 @@
+---
+build_srpm_tito_releaser_args: []
+build_srpm_dist_args: ''
```
evgeni: what is this (and the next one) for? I don't see the vars used.

obal/data/roles/build_srpm/tasks/main.yaml
komidore64: Hrm, yeah that does seem a bit goofy. Let me clean this up.

obal/data/roles/build_srpm/defaults/main.yaml
komidore64: removed

obal/data/roles/build_srpm/tasks/main.yaml
komidore64: defaults to `build_package_tito_releaser_args` but overrides with passed in `--dist` and/or `--scl` flags, or `build_srpm_tito_releaser_args` if you like extra vars.

komidore64:
rebased

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -1,10 +1,36 @@
 ---
+- block:
+    - name: 'Set dist from commandline args'
+      set_fact:
+        build_srpm_tito_releaser_args: "{{ build_srpm_tito_releaser_args | default([]) }} + [ '--dist {{ build_srpm_dist }}' ]"
```
evgeni: given `build_srpm_tito_releaser_args` is defined in `defaults`, I think we don't need the `| default([])` here?

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -1,10 +1,36 @@
...
+      when: build_srpm_dist is defined
+
+    - name: 'Set scl from commandline args'
+      set_fact:
+        build_srpm_tito_releaser_args: "{{ build_srpm_tito_releaser_args | default([]) }} + [ '--scl {{ build_srpm_scl }}' ]"
```
evgeni: given `build_srpm_tito_releaser_args` is defined in `defaults`, I think we don't need the `| default([])` here?

tests/fixtures/mockbin/mockbin
```diff
@@ -45,10 +48,12 @@ if prog == 'tito':
     parser_release.add_argument('-y', action='store_const', dest='output', const='Created task: 1234')
     parser_release.add_argument('--arg')
     parser_build = subparsers.add_parser('build')
-    parser_build.add_argument('--srpm', action='store_const', dest='output', const='Wrote: hello.src.rpm')
+    parser_build.add_argument('--srpm', action='store_const', dest='output', const="Wrote: {}".format(_FAKE_SRPM_PATH))
     parser_build.add_argument('--scl', action='store')
     parser_build.add_argument('--offline', action='store_true')
     args = parser.parse_args()
+    if 'output' in args and re.compile(r'Wrote').search(args.output):
+        open(_FAKE_SRPM_PATH, 'w').close()
```
evgeni: `/tmp/tito` does not exist at this point, so you have to `os.mkdir` it first (and catch the case where it actually does exist, like on all our workstations).

tests/fixtures/mockbin/mockbin
```diff
@@ -45,10 +48,12 @@ if prog == 'tito':
     parser_release.add_argument('-y', action='store_const', dest='output', const='Created task: 1234')
     parser_release.add_argument('--arg')
     parser_build = subparsers.add_parser('build')
-    parser_build.add_argument('--srpm', action='store_const', dest='output', const='Wrote: hello.src.rpm')
+    parser_build.add_argument('--srpm', action='store_const', dest='output', const="Wrote: {}".format(_FAKE_SRPM_PATH))
     parser_build.add_argument('--scl', action='store')
     parser_build.add_argument('--offline', action='store_true')
     args = parser.parse_args()
+    if 'output' in args and re.compile(r'Wrote').search(args.output):
```
evgeni: I think you can just write `if 'output' in args and _FAKE_SRPM_PATH in args.output`

komidore64:
@evgeni rebased and updated

obal/data/playbooks/srpm/metadata.obal.yaml
```diff
@@ -3,3 +3,13 @@ help: >
 
 
   Building source RPMs can be useful when locally building packages.
+variables:
+  build_srpm_dist:
+    parameter: --dist
+    help: Dist to set when building a SRPM
+  build_srpm_scl:
```
ekohl: I thought SRPMs weren't aware of SCL since it's just a combination of source + spec.

komidore64: They technically aren't SCL aware, but tito can still prepend the scl prefix to the front of the SRPM name.

komidore64:
@ekohl removed

ekohl:
@komidore64 I meant to create a PR with only the ansible-lint fixes so we could merge that. This one needs squashing anyway :)

evgeni:
bleh, the travis errors are unrelated, your change looks good

komidore64:
squashed

PR #164, opened 2018-12-14
Add a bumpversion config

ekohl:


evgeni:
ACK

I think we should add "releasing" to the developer docs? :)

PR #168, opened 2019-01-16
pin pytest-xdist to unfuck tests

evgeni:


tox.ini
```diff
@@ -14,7 +14,7 @@ deps =
     pytest-cov
     pytest-flakes
     pytest-pylint
-    pytest-xdist
+    pytest-xdist==1.25.0
```
komidore64: would `< 1.26.0` be better?

evgeni: I don't expect a 1.25.z, but updated :)

komidore64: ah okay

PR #173, opened 2019-02-08
add a workaround for annex not liking github sources

evgeni:
github not always sends the "Content-Length" header when git annex tries
to fetch artifacts from there. some annex versions are more lax about
this (like the one in EPEL7), others (like in Fedora) bail out with:

    while adding a new url to an already annexed file, url does not have expected file size

this commit adds a workaround for this behaviour, by removing the old
annexed file before trying to add it. it is ensured that the newly added
file is identical by checking the result of git diff.

the workaround is disabled by default and needs to be enabled by passing

    setup_sources_annex_workaround=true

### DIFF
--- a/obal/data/roles/setup_sources/tasks/annex.yml
+++ b/obal/data/roles/setup_sources/tasks/annex.yml
@@ -24,6 +24,12 @@
     ( setup_sources_source_exists.stat.exists and setup_sources_source_exists.stat.islnk is defined
     and setup_sources_source_exists.stat.islnk )"
 
+- name: 'Remove source'
+  file:
+    name: "{{ setup_sources_sourcepath }}"
+    state: absent
+  when: setup_sources_search_annex is failed
+
 - name: 'Add url to annex for source'
   command: >-
     git annex addurl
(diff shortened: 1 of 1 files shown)


obal/data/roles/setup_sources/defaults/main.yaml
```diff
@@ -1,2 +1,3 @@
 ---
 setup_sources_annex_options: ""
+setup_sources_annex_workaround: false
```
ehelms: Why would we want this work around to be configurable?

evgeni: I initially thought to make this configurable because it does not happen on the epel version of git-annex that we use on the foreman infra.

But then every user would have to enable that themselves, so you're right and I should drop the flag again.

evgeni: dropped

evgeni:
the ansible-lint error is unrelated

zjhuntin:
Just to be sure, the updated approach isn't to have 
`setup_sources_annex_workaround` set but instead if it does fail due to annex try this method instead?

evgeni:
@zjhuntin nope, it just unconditionally removes the old annexed file, lets annex add it back, and then runs `git diff` to ensure no *actual* change happened but annex is happy now.

pcreech:
> @zjhuntin nope, it just unconditionally removes the old annexed file, lets annex add it back, and then runs `git diff` to ensure no _actual_ change happened but annex is happy now.

This sounds like a rather effective way to ensure referential integrity.  +1

PR #174, opened 2019-02-15
also test on python 3.7

evgeni:


evgeni:
Yes to metadata and also rather no to `python_requires`

PR #176, opened 2019-02-28
only fail annex task when we actually removed the source before

evgeni:
changing the source is legit for nightly packages


evgeni:
https://travis-ci.org/theforeman/obal/builds/499746213 is green, whyever it did not report back properly.

PR #179, opened 2019-03-15
Introduce modules for tito build and release

ekohl:
This isn't tested in depth, but it attempts to clean up the `build_package` role.


obal/data/roles/build_package/tasks/koji.yml
```diff
@@ -1,25 +1,27 @@
 ---
 - name: 'Set tito_releasers for brew scratch'
   set_fact:
-    build_package_tito_releasers: "{{ releasers | join(' ') | replace('dist-git', 'scratch') }}"
+    build_package_tito_releasers: "{{ releasers | join(' ') | replace('dist-git', 'scratch') | split(' ') }}"
```
ekohl: I don't know if there's a better way to do this.

evgeni:
```yaml
{{ releasers | map('replace', 'dist-git', 'scratch') | list }}
```

evgeni:
I like this A LOT!

Could you make Travis happy please? I think you generate the `tito` params slightly differently, and that trips our `mockbin`

obal/data/roles/setup_sources/tasks/git.yml
```diff
@@ -14,4 +14,4 @@
 - name: 'Register source dir tito argument'
   set_fact:
     build_package_tito_releaser_args:
-    - '--arg source_dir={{ setup_sources_git_checkout_dir }}'
+    - 'source_dir={{ setup_sources_git_checkout_dir }}'
```
ekohl: I had some doubts about changing this part. Should we merge a matching change in foreman-packaging at the same time or write compatibility code?

obal/data/playbooks/nightly/nightly.yaml
```diff
@@ -5,6 +5,8 @@
   gather_facts: no
   vars:
     build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args | default([]) }}"
+  roles:
+    - package_variables
```
ekohl: I needed to add this for `package_base_dir` to be defined. Not sure if there's a better way and if it's needed in this PR.

obal/data/modules/tito_build.py
```diff
@@ -0,0 +1,57 @@
...
+            arguments=dict(type='list', required=False),
+            srpm=dict(type='bool', required=False, default=False),
+            offline=dict(type='bool', required=False, default=False),
+            dist=dict(type='str', required=False),
+            scl=dict(rtype='str', equired=False),
+            output=dict(type='path', required=False),
+        )
+    )
+
+    command = ['tito', 'build']
```
ekohl: I thought about factoring out this param parsing code but I can't come up with anything that's more readable

ekohl:
@evgeni thoughts about the duplication and pylint errors that fail the build?

ehelms:
Committed now via another PR!

PR #180, opened 2019-03-18
Build the actual commit in a container

ekohl:
Previously we tested if we could build master, but that's rather useless on PR testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ install:
   - chmod +x /home/travis/.local/bin/rpmdev-packager
 script:
   - tox
-  - docker build --build-arg VERSION=master -t quay.io/foreman/obal:latest .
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,7 +15,7 @@ install:
   - chmod +x /home/travis/.local/bin/rpmdev-packager
 script:
   - tox
-  - docker build --build-arg VERSION=master -t quay.io/foreman/obal:latest .
+  - docker build --build-arg VERSION=$TRAVIS_COMMIT -t quay.io/foreman/obal:latest .
```
ehelms: That doesn't seem to work. Maybe we try HEAD ?

ekohl: The problem is that it's a PR reference which isn't fetched by default. That's why it can't check it out.

evgeni:
This is fixed by moving to GH actions

PR #181, opened 2019-03-19
obal was quietly setting an env variable that debug output didn't show

komidore64:


obal/__init__.py
```diff
@@ -324,7 +324,7 @@ def main(cliargs=None):  # pylint: disable=R0914
     ansible_playbook = (["ansible-playbook"] + ansible_args)
 
     if args.verbose:
-        print(ansible_playbook)
+        print("ANSIBLE_CONFIG={} {}".format(os.environ["ANSIBLE_CONFIG"], ' '.join(ansible_playbook)))
```
ekohl: This is the same, but wondering if it's better.
```suggestion
        print("ANSIBLE_CONFIG={}".format(os.environ["ANSIBLE_CONFIG"]), ' '.join(ansible_playbook))
```

komidore64:
PR title had to stick with the obfuscated theme :stuck_out_tongue:

obal/__init__.py
ekohl: Note that I think this is just enough to get you under a 100 characters and make pylint happy.

komidore64: Awesome!

komidore64:
I fixed Travis's `ansible-lint` problem in #182 by pinning `PyYaml` to `< 5.1`.

PR #182, opened 2019-03-19
DRY up finding the spec file

komidore64:


obal/data/roles/spec_file/meta/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+dependencies:
```
ekohl: I think this also depends on `package_variables` to define `package_base_dir`.

komidore64: `ensure_package` depends on `package_variables` so I didn't feel the need to specify it here also.

obal/data/roles/setup_sources/tasks/specfile.yml
```diff
@@ -1,16 +1,10 @@
...
-- name: get spec file
-  find:
-    pattern: "*.spec"
-    path: "{{ package_dir }}"
-  register: setup_sources_spec_file
-
 - name: 'Set spec file path'
   set_fact:
-    setup_sources_spec_file_path: "{{ setup_sources_spec_file.files[0].path }}"
+    spec_file_path: "{{ spec_file.files[0].path }}"
```
ekohl: This is only used in one place. Perhaps just use `spec_file.files[0].path` in the `spectool` command below?

obal/data/roles/diff_package/tasks/git_package_info.yml
```diff
@@ -1,10 +1,4 @@
 ---
-- name: get spec file
-  find:
-    pattern: "*.spec"
-    path: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
-  register: spec_file
-
 - name: "Set spec file path, scl options and releasers"
```
ekohl: This looks outdated as well.

komidore64: Oh it sure is!

komidore64:
> Given the repetition you may also consider setting `spec_file_path` in the `spec_file` role.

This is a good idea. I decided to set `spec_file_path` in the `spec_file` role.

obal/data/roles/diff_package/tasks/git_package_info.yml
komidore64: fixed

obal/data/roles/setup_sources/tasks/specfile.yml
komidore64: I decided to move `spec_file_path` assignment into the `spec_file` role.

evgeni:
> Thought experiment: should we become opinionated and prefer the spec filename to match the directory name (as the Fedora packaging guidelines suggest)? We now have an easy place to start enforcing it.

I'd prefer not to do this (yet), as that would require changes to all our downstream packaging repos.

komidore64:
> I'd prefer not to do this (yet), as that would require changes to all our downstream packaging repos.

You beat me to saying this!

obal/data/playbooks/changelog/changelog.yaml
```diff
@@ -4,14 +4,8 @@
...
   tasks:
-    - name: get spec file
-      find:
-        pattern: "*.spec"
-        path: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
-      register: spec_file
-
     - name: 'ensure changelog'
       changelog:
         spec: "{{ spec_file.files[0].path }}"
```
evgeni:
```suggestion
        spec: "{{ spec_file_path }}"
```

obal/data/playbooks/bump-release/bump-release.yaml
```diff
@@ -6,14 +6,8 @@
...
+    - spec_file
   tasks:
-    - name: get spec file
-      find:
-        pattern: "*.spec"
-        path: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
-      register: spec_file
-
     - name: 'Bump release'
       command: "rpmdev-bumpspec {{ spec_file.files[0].path | basename }} -c \"{{ changelog }}\""
```
evgeni:
```suggestion
      command: "rpmdev-bumpspec {{ spec_file_path | basename }} -c \"{{ changelog }}\""
```

komidore64:
@evgeni your suggestions have been incorporated!

ekohl:
> I'd prefer not to do this (yet), as that would require changes to all our downstream packaging repos.

We could start with a configurable warning that could be an error. That would help us make sure foreman-packaging's rpm/develop doesn't regress.

komidore64:
I'm not sure why this blew up.

PR #183, opened 2019-03-21
new module - package_whitelist_check

komidore64:


komidore64:
this is probably going to fail due to `PyYaml` just like #181 and #182.

obal/__init__.py
```diff
@@ -321,7 +321,11 @@ def main(cliargs=None):  # pylint: disable=R0914
     from ansible.cli.playbook import PlaybookCLI
 
     ansible_args = generate_ansible_args(inventory_path, args, parser.obal_arguments)
-    ansible_playbook = (["ansible-playbook"] + ansible_args)
+    ansible_playbook = (
+        ["ansible-playbook"]
+        + ["-e", "ansible_python_interpreter='/usr/bin/env python3'"]
```
ehelms: Does this degrade gracefully for systems that only have python2?

obal/data/module_utils/obal.py
```diff
@@ -11,20 +27,31 @@ def get_changelog_evr(specfile):
...
+
+def get_specfile_name(specfile):
+    """get the name from the specfile"""
+    return specfile_macro_lookup(specfile, '%{name}')
+
+def get_whitelist_status(build_command, tag, package):
+    """
+    Get whitelist status of a given package within a tag.
+
+    Return `True` if the package is whitelistes, `False` otherwise.
```
ehelms: s/whitelistes/whitelisted

obal/__init__.py
ekohl: No. Ansible 2.8 will have autodetection for this so it will no longer be needed, but this breaks EL7 support.

obal/data/roles/build_package/tasks/koji.yml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: "Confirm package is whitelisted"
+  package_whitelist_check:
+    releasers_conf: "{{ inventory_dir }}/rel-eng/releasers.conf"
```
ehelms: Why do we need a releasers.conf if we have releasers specified?

obal/__init__.py
komidore64: I think it's about time that obal decides to support either py2 or py3. @ehelms, @ekohl thoughts?

obal/data/module_utils/obal.py
komidore64: whoops, thank you.

obal/__init__.py
ekohl: :-1: it means we can no longer run obal on EL7 machine. In the Foreman infra we currently don't have any Fedora slaves and I'd like to avoid installing them just to run obal. However, I still want to use obal on my Fedora machine.

obal/data/module_utils/obal.py
komidore64: fixed

obal/data/roles/build_package/tasks/koji.yml
komidore64: I need both the given releaser and `releasers.conf` so i can figuring out how a releaser is mapped to a tag.

komidore64:
@ekohl i made the module py2 compatible.

obal/data/module_utils/obal.py
```diff
@@ -16,15 +32,25 @@ def get_changelog_evr(specfile):
...
+    cmd = [
+        build_command,
+        'list-pkgs',
+        '--tag',
+        tag,
+        '--package',
+        package,
+        '--quiet'
+    ]
+    return not bool(subprocess.call(cmd))
```
evgeni: Clever, but I'd prefer
```python
retcode = subprocess.call(cmd)
return retcode == 0
```
as I think this is more readable?

obal/data/roles/build_package/meta/main.yml
```diff
@@ -1,6 +1,4 @@
 ---
 dependencies:
   - { role: setup_workspace }
-  - { role: ensure_package }
   - { role: setup_sources }
-  - { role: package_variables }
```
evgeni: these two are not required anymore after your cleanup in #182, right?

obal/data/roles/build_package/tasks/koji.yml
```diff
@@ -1,4 +1,11 @@
...
+    spec_file_path: "{{ spec_file.files[0].path }}"
```
evgeni: I guess this can be `spec_file_path` now?

https://github.com/theforeman/obal/blob/f2dcda2314eb8e84b1cae69c65b201df192c1248/obal/data/roles/spec_file/tasks/main.yml#L10

obal/data/module_utils/obal.py
```diff
@@ -3,6 +3,22 @@
 """
 import subprocess
 
+def specfile_macro_lookup(specfile, macro_str):
```
ekohl: Nit: coding style wants 2 empty lines before a new function

obal/data/module_utils/obal.py
```diff
@@ -3,6 +3,22 @@
...
+    """expand a given macro from a specfile"""
+    subproc_output = subprocess.check_output([
```
ekohl: Nit: I think the `command = [\n...\n]\nreturn subprocess.check_output(command)` format might be slightly easier to read.

obal/data/modules/package_whitelist_check.py
```diff
@@ -0,0 +1,75 @@
...
+    # fail if we don't find any branches or autobuild_tags
+    if (result['branches'] + result['autobuild_tags']) is []:
+        module.fail_json(msg="No branches or autobuild_tags were found mapped to {}.".format(module.params['releasers']), **result)
+
+    # check whitelist status
+    for tag in (result['branches'] + result['autobuild_tags']):
+        status = get_whitelist_status(module.params['build_command'], tag, package_name)
+        result['whitelist_status'][tag] = status
+
+    result['all_whitelisted'] = reduce(lambda col, val: col and val, list(result['whitelist_status'].values()))
```
ekohl: Is this equal to: `result['all_whitelisted'] = all(result['whitelist_status'].values())`

obal/data/modules/package_whitelist_check.py
```diff
@@ -0,0 +1,75 @@
...
+
+    package_name = get_specfile_name(module.params['spec_file_path'])
+
+    # get branches and tags
+    for releaser in module.params['releasers']:
+        if config.has_option(releaser, 'branches'): result['branches'] = config.get(releaser, 'branches').split(' ')
+        if config.has_option(releaser, 'autobuild_tags'): result['autobuild_tags'] = config.get(releaser, 'autobuild_tags').split(' ')
+
+    # fail if we don't find any branches or autobuild_tags
+    if (result['branches'] + result['autobuild_tags']) is []:
```
ekohl: I'd not check with `is` since that checks the memory location. Probably better to use `if not result['branches'] and not result['autobuild_tags']:`

obal/__init__.py
komidore64: i removed this section where `ansible_python_interpreter` was being set.

obal/data/module_utils/obal.py
komidore64: Your suggestion is clearer.

obal/data/roles/build_package/meta/main.yml
komidore64: They are still depended upon, but it's not necessary to specify them.

Role dependency resolves as follows:

```
build_package
|-- setup_workspace
 `- setup_sources
    |-- git_annex_setup
    |    `- package_variables
     `- spec_file
         `- ensure_package
             `- package_variables
```

We don't need to specify `package_variables` or `ensure_package` since we've already got `setup_sources`.

obal/data/roles/build_package/tasks/koji.yml
komidore64: You are correct!

obal/data/module_utils/obal.py
komidore64: ack

obal/data/modules/package_whitelist_check.py
komidore64: Ah, I'll remember `is` one of these days ...

obal/data/modules/package_whitelist_check.py
komidore64: Yes, you're correct. This looks much cleaner, thank you.

obal/data/module_utils/obal.py
komidore64: Refactoring.

komidore64:
@evgeni, @ekohl pull request updated based on feedback and suggestions.

komidore64:
Once I get some :+1:, I'd prefer to reorganize my commits before merging, thanks!

obal/data/modules/package_whitelist_check.py
```diff
@@ -0,0 +1,74 @@
...
+
+    package_name = get_specfile_name(module.params['spec_file_path'])
+
+    # get branches and tags
+    for releaser in module.params['releasers']:
+        if config.has_option(releaser, 'branches'): result['branches'] = config.get(releaser, 'branches').split(' ')
+        if config.has_option(releaser, 'autobuild_tags'): result['autobuild_tags'] = config.get(releaser, 'autobuild_tags').split(' ')
+
+    # fail if we don't find any branches or autobuild_tags
+    if not result['branches'] and not result['autobuild_tags']):
```
ekohl:
```suggestion
    if not result['branches'] and not result['autobuild_tags']:
```

obal/data/modules/package_whitelist_check.py
```diff
@@ -0,0 +1,74 @@
...
+        module.fail_json(msg="Config file [ {} ] could not be found.".format(releasers_config), **result)
+
+    config = configparser.ConfigParser(allow_no_value=True)
+    config.read(releasers_config)
+
+    package_name = get_specfile_name(module.params['spec_file_path'])
+
+    # get branches and tags
+    for releaser in module.params['releasers']:
+        if config.has_option(releaser, 'branches'): result['branches'] = config.get(releaser, 'branches').split(' ')
```
ekohl: I'd add a new line after `:`. Also note that you're conditionally setting the parameter which means using it as `result['branches']` will raise a `KeyError`.

obal/data/modules/package_whitelist_check.py
```diff
@@ -0,0 +1,74 @@
...
+    # fail if we don't find any branches or autobuild_tags
+    if not result['branches'] and not result['autobuild_tags']):
+        module.fail_json(msg="No branches or autobuild_tags were found mapped to {}.".format(module.params['releasers']), **result)
+
+    # check whitelist status
+    for tag in (result['branches'] + result['autobuild_tags']):
+        status = get_whitelist_status(module.params['build_command'], tag, package_name)
+        result['whitelist_status'][tag] = status
+
+    result['all_whitelisted'] = all(result['whitelist_status'].values()))
```
ekohl: Does storing it in the result make sense? Either the command failed (which you can check in another way) or it's true.

obal/data/modules/package_whitelist_check.py
komidore64: Oh no I missed a thing!

obal/data/modules/package_whitelist_check.py
komidore64: `result['branches']` is set to a default of `[]` above in an attempt to prevent that from occurring.

obal/data/modules/package_whitelist_check.py
komidore64: Yeah, I wasn't particularly sure whether it was worth storing or not, so I just threw it in there.

komidore64: I decided to remove it.

obal/data/modules/package_whitelist_check.py
komidore64: newlines added

ekohl: Right, I missed that :)

komidore64:
commits squashed and travis should _finally_ be green!

obal/data/roles/build_package/meta/main.yml
evgeni: I would also have believed you, if you just wrote "yes" ;)

obal/data/modules/rpmspec_changelog_check.py
```diff
@@ -10,16 +10,11 @@
 from ansible.module_utils.obal import get_changelog_evr, get_specfile_evr  # pylint:disable=import-error,no-name-in-module
 
 
-ANSIBLE_METADATA = {
```
ekohl: What is the reason to remove this?

komidore64: It doesn't seem like it's used anywhere, and pylint was failing it on a `duplicate-code` error that I was unable to disable.

evgeni: As long as this module is not part of ansible proper or we're trying to generate docs using the ansible sphinx (as we do for f-a-m), this data should have no use.

Simple test, if `ansible-doc rpmspec_changelog_check` still works after this change, we're good :)

komidore64:
```
$ ansible-doc -M ./ rpmspec_changelog_check
ERROR! module rpmspec_changelog_check missing documentation (or could not parse documentation): 'NoneType' object has no attribute 'get'
```

PR #184, opened 2019-03-27
use a non-broken rpm-bumpspec

evgeni:


komidore64:
Would this be a good time to set those other executables to a specific hash rather than `master` also? avoid something like this in the future?

evgeni:
I've been thinking about that too, yeah. If only pagure could reference a tag like if it's a branch (a specific commit feels ugly, so I'd pin to the last known release).

Anyways, I can revisit that if you like to

komidore64:
meh, it's fine.

PR #186, opened 2019-04-02
repoclosure based on dist instead of package

evgeni:
this PR changes how we handle repoclosure in obal, by splitting the two usecases we have:

1. when acting on a set of downloaded rpms, decide which repoclosure set (lookasides and targets) to look at by adding the subfolder of the downloaded rpms to the search path
  for this to work, the scratch command now downloads rpms into `downloaded_rpms/%{dist}` instead of `downloaded_rpms`, where `%{dist}` is the last token of the build-target
1. when acting only on real repos, there is no `dist` to look for, which makes the code a tad ugly at the moment

marking as WIP for now, as merging this will require changes to the `package_manifest`s of the related projects, which I did not prepare yet.


obal/data/roles/repoclosure/tasks/dist.yml
```diff
@@ -0,0 +1,18 @@
+- name: 'Run repoclosure for {{ item }}'
```
ehelms: Should we use something other than `item` thats a bit more clear? and then use a `loop_var` to set this on the task using this?

obal/data/roles/repoclosure/tasks/regular.yml
```diff
@@ -0,0 +1,17 @@
+- name: 'Run repoclosure'
```
ehelms: Remind me what makes it regular vs. irregular?

obal/data/roles/repoclosure/tasks/dist.yml
evgeni: sure can, good idea

obal/data/roles/repoclosure/tasks/regular.yml
evgeni: `irregular` = `dist` = "we downloaded a set of RPMs and test them WITH a set of repos"
`regular` = "here are the repos, run repoclosure"

Totally taking better names (or, more probably, will try to make those two files one).

evgeni:
Copying a comment from @ekohl in https://github.com/theforeman/foreman-packaging/pull/3585 so we don't loose it:

"Does this mean we no longer check repoclosure for nonscl without plugins present? Should we have a separate run for that so you never break repoclosure on nightly, but can break plugins and katello. For example when you're doing the fog migration."

This is related to the fact that I've added plugins and katello to "target" when running repoclosure for core, so that we can verify that "core" didn't break "plugins" or "katello". But it obviously has the downside that now both are also used for dependency resolution of "core", which is not what we want.

ehelms:
Maybe we want that? Or maybe thats a CI nuance where we want core to warn about katello and plugins but not fail on it. "Hey you are breaking plugins! proceed with caution"

evgeni:
The packaging repo is currently not set up to block a merge when CI is red. So if we report "core closure: green, plugins closure: red, katello closure: red", the reviewer can still merge it (preferably with a comment "yes, I know it breaks katello, fixes are on their way"). However, this still requires us to run three sets of repoclosures for one package, which is something our infra is currently not set up for.

What we'd need is (for each package group) as list of "affected" repositories, and then run a repoclosure for each of these affected once.

obal/data/roles/repoclosure/tasks/regular.yml
evgeni: I made it one now

ehelms:
What else are you planning for this to move out of draft state?

obal/data/roles/repoclosure/defaults/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
 repoclosure_additional_repos:  []
 repoclosure_config: repoclosure/el7.conf
-repoclosure_target_repos: [ "{{ repoclosure_target_repo | default('downloaded_rpms') }}" ]
+repoclosure_target_repos: ['downloaded_rpms']
```
evgeni: I wonder if the old behaviour is better. But this is less magic.

evgeni:
@ehelms my last commit and the prep-work for the changes to f-packaging (which are now there)

PR #189, opened 2019-04-10
koji_whitelist_check - add package- and branch-name flexibility

komidore64:


ehelms:
Does this require changes in packaging repositories to use this functionality or will it just work?

komidore64:
> Does this require changes in packaging repositories to use this functionality or will it just work?

it will require new fields added to `package_manifest.yaml` for packages that don't have one-to-one naming for branches<->tags or for packages within an SCL.

ehelms:
Would you mind providing an example comment of what, for example, foreman-packaging would need added to it ?I am having a hard time wrapping my mind around it.

komidore64:
take `rubygem-katello` for example:

```
# this is what gets run
koji list-pkgs --tag katello-nightly-rhel7 --package rubygem-katello
# => no known package

# what we want
koji list-pkgs --tag katello-nightly-rhel7 --package tfm-rubygem-katello
```

we can alter the `list-pkgs` lookup with a `koji_whitelist_package_replace` to its inventory entry:

```
# package_manifest.yaml

rubygem-katello:
  koji_whitelist_package_replace:
    regexp: '^'
    replace: 'tfm-'
```

komidore64:
The same mechanism works with `koji_whitelist_branch_replace` for downstream cases were the dist-git branch name doesn't match the corresponding tag character-for-character.

komidore64:
@ekohl that script seems useful, perhaps it should be moved into obal proper so that other packaging repositories may benefit as opposed to solely `foreman-packaging`?

ehelms:
This is pretty explicit. I had wondered about pulling from tito (though it ties things to a tito config) or being able to specify a parameter like `scl: tfm-` or `package_prefix: tfm-` to prepend.

komidore64:
it looks like obal looks for a `scl` host variable in some places. I'll see about using that instead.

https://github.com/theforeman/obal/blob/23a843a42d4dc5936a09e09ca0824d7045086348/obal/data/roles/diff_package/tasks/git_package_info.yml#L7

komidore64:
@ehelms yeah i'd prefer to rely on tito as little as possible if we can help it.

komidore64:
@ekohl would you mind expanding on your opposition?

ekohl:
My opposition is about introduction of additional variables. These are not as straight forward as you might think and can really complicate release management. Usually because they can get out of sync.

The reality in foreman-packaging is currently that Tito sets the `scl_prefix` variable for some tags and a spec file can (but doesn't have to) use this. Currently this is only applied in a subset of `rubygem-*` packages.

Some edge cases to be aware of (in foreman-packaging):
* We have some packages which are built into two tags, one SCL and the other non-SCL. See the `foreman_scl_and_nonscl_packages` group in `package_manifest.yaml`.
* We have some some packages which we've duplicated because the version differs in SCL and non-SCL. See (tfm-)rubygem-jwt and (tfm-)rubygem-rest-client).

In [compare-with-tags.rb](https://github.com/theforeman/foreman-packaging/blob/4c03949229e518e777a94ba4d1f3dec470b852cc/rel-eng/compare-with-tags.rb#L13-L15) we do use the `scl` value from tito.props. It may be possible to loop over `diff_package_tags` and determine the value for each one.

PR #191, opened 2019-04-23
ability to force a scratch build

komidore64:


obal/data/roles/build_package/defaults/main.yaml
```diff
@@ -9,3 +9,4 @@ build_package_wait: true
 build_package_download_logs: false
 build_package_download_rpms: false
 build_package_waitrepo: true
+run_koji_whitelist_check: true
```
evgeni: We've been prefixing role-specific vars with the role name, so this would become `build_package_run_koji_whitelist_check` or `build_pacakge_koji_whitelist_check`

obal/data/playbooks/scratch/metadata.obal.yaml
```diff
@@ -2,3 +2,8 @@ help: |
   Create a scratch build of a package
 
   A scratch build produces the same result as a normal release would, but it's not tagged in any repository. This allows verification of a change without actually committing it.
+variables:
+  run_koji_whitelist_check:
+    action: store_false
+    parameter: --force
```
evgeni: `force` sounds very generic to me and I'd prefer `skip-koji-check` or `ignore-koji-check` or something else that makes the parameter closer to what's actually skipped.

komidore64:
@evgeni both of these are great suggestions! thanks. I'll see if I can get those in today.

komidore64:
@evgeni updated with your suggestions!

PR #199, opened 2019-07-05
Add GitSourceBuilder fetch strategy for scratch builds

pcreech:
Also expose cli arges for use with scratch building

### DIFF
--- a/obal/tito_extensions/git_source_strategy.py
+++ b/obal/tito_extensions/git_source_strategy.py
@@ -1,0 +1,183 @@
+# Copyright (c) 2008-2014 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 2 (GPLv2). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+# along with this software; if not, see
+# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+#
+# Red Hat trademarks are not licensed under GPLv2. No permission is
+# granted to use or replicate Red Hat trademarks that are incorporated
+# in this software or its documentation.
+
+from datetime import datetime
+import glob
+import json
+import os
+import os.path
+import re
+import shutil
+import subprocess
+from zipfile import ZipFile
+
+from tito.builder.fetch import SourceStrategy
+from tito.common import error_out, debug, run_command
+
+class GitSourceStrategy(SourceStrategy):
+    """
+    Designed to be used for nightly or pull request builds only.
+    It first copies source files from git, then copies the source file(s)
+    over the top, so they're merged (patches etc can then be stored
+    in git).
+    """
+    def fetch(self):
+        if "source_dir" not in self.builder.args:
+            raise Exception("Specify '--arg source_dir=...'")
+
+        # Copy the live spec from our starting location. Unlike most builders,
+        # we are not using a copy from a past git commit.
+        self.spec_file = os.path.join(self.builder.rpmbuild_sourcedir,
+                    '%s.spec' % self.builder.project_name)
+        shutil.copyfile(
+            os.path.join(self.builder.start_dir, '%s.spec' %
+                self.builder.project_name),
+            self.spec_file)
+
+        gitrev = self._fetch_local()
+
+        for s in os.listdir(self.builder.start_dir):
+            if os.path.exists(os.path.join(self.builder.start_dir, s)):
+                shutil.copyfile(
+                    os.path.join(self.builder.start_dir, s),
+                    os.path.join(self.builder.rpmbuild_sourcedir, os.path.basename(s)))
+        print("  %s.spec" % self.builder.project_name)
+
+        replacements = []
+        src_files = run_command("find %s -type f" %
+              os.path.join(self.builder.rpmbuild_sourcedir, 'archive')).split("\n")
+
+        def filter_archives(path):
+            base_name = os.path.basename(path)
+            return ".tar" in base_name
+
+        for i, s in enumerate(filter(filter_archives, src_files)):
+            base_name = os.path.basename(s)
+            debug("Downloaded file %s" % base_name)
+
+            dest_filepath = os.path.join(self.builder.rpmbuild_sourcedir,
+                    base_name)
+            shutil.move(s, dest_filepath)
+            self.sources.append(dest_filepath)
+
+            # Add a line to replace in the spec for each source:
+            source_regex = re.compile("^(source%s:\s*)(.+)$" % i, re.IGNORECASE)
+            new_line = "Source%s: %s\n" % (i, base_name)
+            replacements.append((source_regex, new_line))
+
+        # Replace version in spec:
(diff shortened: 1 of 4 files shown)


ekohl:
Some quick notes that we had during an offline discussion. The mutli package flow where you pass in `obal scratch foo bar --git-{something} blah` will be an issue. @pcreech suggested variable replacement so it would be `obal scratch foo bar --git-source-dir "checkouts/{{ inventory_name }}"`. It's untested, but might work.

obal/tito_extensions/git_source_strategy.py
```diff
@@ -0,0 +1,183 @@
...
+
+        replacements = []
+        src_files = run_command("find %s -type f" %
+              os.path.join(self.builder.rpmbuild_sourcedir, 'archive')).split("\n")
+
+        def filter_archives(path):
+            base_name = os.path.basename(path)
+            return ".tar" in base_name
+
+        for i, s in enumerate(filter(filter_archives, src_files)):
```
ekohl: Maybe we can drop the additional function:

```suggestion
        for i, s in enumerate(s for s in src_files if ".tar" in os.path.basename(s)):
```

obal/tito_extensions/git_source_strategy.py
```diff
@@ -0,0 +1,183 @@
...
+            self.sources.append(dest_filepath)
+
+            # Add a line to replace in the spec for each source:
+            source_regex = re.compile("^(source%s:\s*)(.+)$" % i, re.IGNORECASE)
+            new_line = "Source%s: %s\n" % (i, base_name)
+            replacements.append((source_regex, new_line))
+
+        # Replace version in spec:
+        version_regex = re.compile("^(version:\s*)(.+)$", re.IGNORECASE)
+        self.version = self._get_version()
```
ekohl: Is this `self.version` used by anything else or could it be a local variable? There's a few other instances that I didn't check

obal/tito_extensions/git_source_strategy.py
```diff
@@ -0,0 +1,183 @@
...
+        print("Building version: %s" % self.version)
+        replacements.append((version_regex, "Version: %s\n" % self.version))
+        self.replace_in_spec(replacements)
+
+        rel_date = datetime.utcnow().strftime("%Y%m%d%H%M")
+        self.release = rel_date + gitrev
+        print("Building release: %s" % self.release)
+        run_command("sed -i '/^Release:/ s/%%/.%s%%/' %s" % (self.release, self.spec_file))
```
ekohl: Should this also use `self.replace_in_spec`?

obal/tito_extensions/git_source_strategy.py
```diff
@@ -0,0 +1,183 @@
...
+
+        fetchdir = os.path.join(self.builder.rpmbuild_sourcedir, 'archive')
+        if not os.path.exists(fetchdir):
+          os.mkdir(fetchdir)
+
+        arch_prefix = "-".join([self.builder.project_name, version])
+
+        # git archive --format=tar.gz --prefix=pulp-2.15.0/ master > pulp-2.15.0.tar.gz
+
+        with open("./%s.tar.gz" % arch_prefix, "w+") as archive:
```
ekohl: Foreman typically uses `.bz2`. The `.tar.gz` is also repeated a few times. Perhaps this should be its own separate variable? The `./` is also not needed.

obal/tito_extensions/git_source_strategy.py
```diff
@@ -0,0 +1,183 @@
...
+            subprocess.call(["git", "archive", "--format=tar.gz", ("--prefix=%s/" % arch_prefix), "HEAD"], stdout=archive)
```
ekohl:
```suggestion
            subprocess.check_call(["git", "archive", "--format", "tar.gz", "--prefix", "%s/" % arch_prefix, "HEAD"], stdout=archive)
```

obal/tito_extensions/git_source_strategy.py
```diff
@@ -0,0 +1,183 @@
...
+
+        sources = glob.glob("./*.tar.gz")
```
ekohl: We already know the filename because we opened it previously. Is this really needed?

obal/tito_extensions/git_source_strategy.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Get the version from the builder.
+        Sources are configured at this point.
+        """
+        # Assuming source0 is a tar.gz we can extract a version from:
+        base_name = os.path.basename(self.sources[0])
+        debug("Extracting version from: %s" % base_name)
+
+        # Example filename: tito-0.4.18.tar.gz:
+        simple_version_re = re.compile(".*-(.*).(tar.gz|tgz|zip|tar.bz2|gem)")
```
ekohl:
```suggestion
        simple_version_re = re.compile(".*-(.*).(tar.(gz|bz2|xz)|tgz|zip|gem)")
```

obal/tito_extensions/git_source_strategy.py
pcreech: GitSourceStrategy inherits from SourceStrategy, which explicitly defines self.version.

this version variable is utilized by tito's mechanisms in various ways.

https://github.com/dgoodwin/tito/blob/master/src/tito/builder/fetch.py#L79

obal/tito_extensions/git_source_strategy.py
pcreech: This is very dependent on what git archive supports as the "format" call.  I.E. you can't specify bz2, as git archive only supports a limited amount of formats.  (tar, tgz, tar.gz, zip).  This is explicitly utilizing git in a git checkout to generate the archive.

As far as `./` , while it is not needed, it is a little more visually explicit in what is going on.  Can change it though if it's strongly preferred.

obal/tito_extensions/git_source_strategy.py
pcreech: Probably not.  I got the initial lines from `foreman-packaging/rel-eng/lib/custom.py`, and didn't think much further about them when adding in the git archive creation above.

obal/tito_extensions/git_source_strategy.py
pcreech: This line is pulled again from the aforementioned `custom.py`.  We can change it, but unchanged it has endured for a long period of time there without needing modification.  I'm hesitant to induce such a change without strong need.

obal/tito_extensions/git_source_strategy.py
pcreech: Possibly.  It may be cleaner.  This was slightly modified from the following block of code to replace the release macro instead:

https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rel-eng/custom/custom.py#L88-L91

obal/tito_extensions/git_source_strategy.py
pcreech: This code was left unchanged from the original custom.py I utilized as a basis:

https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rel-eng/custom/custom.py#L74

It has endured without change for quite some time, and I'm hesitant to change it without a driving breakage.

ehelms:
Looks to need some ignores from lint and test updates.

PR #201, opened 2019-07-22
Introduce an ApplicationConfig class

ekohl:
The goal of the layout is to bundle all the methods that describe where directories or files can be found. This is passed into the main function which means other tools can easily override this.

This is untested and could use tests.

### DIFF
--- a/obal/__init__.py
+++ b/obal/__init__.py
@@ -164,6 +164,48 @@ class Playbook(object):
         return NotImplemented
 
 
+class ApplicationConfig(object):
+    """
+    A class describing the where to find various files
+    """
+
+    @staticmethod
+    def name():
+        """
+        Return the name as shown to the user in the ArgumentParser
+        """
+        return 'obal'
+
+    @staticmethod
+    def data_path():
+        """
+        Return the data path. Houses playbooks and configs.
+        """
+        return os.environ.get('OBAL_DATA', resource_filename(__name__, 'data'))
+
+    @property
+    @staticmethod
+    def inventory_path():
+        """
+        Return the inventory path
+        """
+        return os.environ.get('OBAL_INVENTORY', os.path.join(os.getcwd(), 'package_manifest.yaml'))
+
+    @classmethod
+    def playbooks_path(cls):
+        """
+        Return the default playbooks path
+        """
+        return os.environ.get('OBAL_PLAYBOOKS', os.path.join(cls.data_path, 'playbooks'))
+
+    @classmethod
+    def ansible_config_path(cls):
+        """
+        Return the ansible.cfg path
+        """
+        return os.environ.get('OBAL_ANSIBLE_CFG', os.path.join(cls.data_path, 'ansible.cfg'))
+
+
 def find_playbooks(playbooks_path):
     """
     Find all playbooks in the given path.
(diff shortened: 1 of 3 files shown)


obal/__init__.py
```diff
@@ -298,11 +310,11 @@ def generate_ansible_args(inventory_path, args, obal_arguments):
     return ansible_args
 
 
-def main(cliargs=None):  # pylint: disable=R0914
+def main(cliargs=None, layout=Layout):  # pylint: disable=R0914
     """
     Main command
     """
-    cfg_path = get_ansible_config_path()
+    cfg_path = layout.ansible_config_path
```
ekohl: I was thinking whether this should incorporate the if exists logic, but if you need to create it, this can be useful.

obal/__init__.py
```diff
@@ -164,41 +164,53 @@ def __lt__(self, other):
         return NotImplemented
 
 
-def find_playbooks(playbooks_path):
+class Layout(object):
     """
-    Find all playbooks in the given path.
+    A class describing the where to find various files
```
evgeni: So one thing why I called this `Config` on IRC was: this class could also contain the CLI name of the tool to be passed in

```python
parser = argparse.ArgumentParser('obal')
```

ekohl: That's a good argument. Now I'm thinking about `Application`, but that would imply there's much more logic to it. How about `ApplicationConfig`?

evgeni: Sold.

ekohl: Updated

evgeni:
@ekohl you need to adjust tests :)

ekohl:
It turns out combining `@property` and `@classmethod` doesn't work. Doesn't change the design, just some details in the implementation.

ekohl:
Rebased. It now uses methods rather than properties. It's not the prettiest, but it works.

tests/test_playbooks.py
```diff
@@ -14,7 +14,7 @@ def playbook_id(fixture_value):
     return fixture_value.name
 
 
-@pytest.fixture(params=obal.find_playbooks(obal.get_playbooks_path()), ids=playbook_id)
+@pytest.fixture(params=obal.find_playbooks(obal.ApplicationConfig.playbooks_path), ids=playbook_id)
```
evgeni:
```suggestion
@pytest.fixture(params=obal.find_playbooks(obal.ApplicationConfig.playbooks_path()), ids=playbook_id)
```

obal/__init__.py
```diff
@@ -164,41 +164,55 @@ def __lt__(self, other):
...
-    """
-    Return the default playbooks path
-    """
-    return os.environ.get('OBAL_PLAYBOOKS', os.path.join(_get_data_path(), 'playbooks'))
+    @classmethod
+    def ansible_config_path(cls):
+        """
+        Return the ansible.cfg path
+        """
+        return os.environ.get('OBAL_ANSIBLE_CFG', os.path.join(cls.data_path, 'ansible.cfg'))
```
evgeni:
```suggestion
        return os.environ.get('OBAL_ANSIBLE_CFG', os.path.join(cls.data_path(), 'ansible.cfg'))
```

obal/__init__.py
```diff
@@ -164,41 +164,55 @@ def __lt__(self, other):
...
+        Return the inventory path
+        """
+        return os.environ.get('OBAL_INVENTORY', os.path.join(os.getcwd(), 'package_manifest.yaml'))
 
+    @classmethod
+    def playbooks_path(cls):
+        """
+        Return the default playbooks path
+        """
+        return os.environ.get('OBAL_PLAYBOOKS', os.path.join(cls.data_path, 'playbooks'))
```
evgeni:
```suggestion
        return os.environ.get('OBAL_PLAYBOOKS', os.path.join(cls.data_path(), 'playbooks'))
```

obal/__init__.py
```diff
@@ -164,41 +164,55 @@ def __lt__(self, other):
...
+        Return the data path. Houses playbooks and configs.
+        """
+        return os.environ.get('OBAL_DATA', resource_filename(__name__, 'data'))
 
-def get_inventory_path():
-    """
-    Return the inventory path
-    """
-    return os.environ.get('OBAL_INVENTORY', os.path.join(os.getcwd(), 'package_manifest.yaml'))
+    @property
```
evgeni:
```suggestion
```

PR #202, opened 2019-09-15
Add tito_release and tito_build modules

ehelms:

PR #203, opened 2019-09-15
Add mock

ehelms:


ehelms:
Requires #202

ehelms:
This will also needs tests added. I first wanted to vet the design approach.

ehelms:
This would allow building a single package at a time with mock. This does not solve the concept of recursively building a group of packages until either they all pass or some number of iterations has occurred. That model is harder to do. Mock technically allow doing this through its recurse flag. However, our actions operate on a host at a time. We would need some way to say build all these SRPMs and then run mock recurse on the directory for just a single play and host. The answer might be to use multiple plays in the playbook with the last play being to use `run_once`. I am open to other ideas.

ekohl:
> The answer might be to use multiple plays in the playbook with the last play being to use `run_once`. I am open to other ideas.

Are you saying you'd write source RPMs into a directory then run mock(chain) with `run_once` on it?

ehelms:
> > The answer might be to use multiple plays in the playbook with the last play being to use `run_once`. I am open to other ideas.
> 
> Are you saying you'd write source RPMs into a directory then run mock(chain) with `run_once` on it?

That's my (untested) theory on an approach. I would attempt this natively with Ansible, but I can't see a way to retry an entire playbook X number of times or until a condition is met. Another approach is leave mock as is, single package builder, and leave it up to the user to wrap calling `obal mock` in their own loop using their own wrapper tooling.

ekohl:
I actually wouldn't mind if it's two separate steps: `obal srpm packages && obal mock el7` makes sense to me. Similarly you could also do `obal mock el7 el8` at some point to build those in parallel.

ehelms:
I'm less clear on what `mock el7` would be doing. I have a guess. However, my first question would be, based on your comment how would I:

 * mock build a single package
 * mock build a group of packages (e.g. foreman or foreman_installer + dependencies)

ekohl:
I was thinking that it would run `SRPMs/*` through the EL7 mock config. Currently we have two so you'd have el7-scl and `el7-nonscl` but if we're going SCL-only that would simplify to `el7`. For something like the client release RPM you should also build for `f29` for example.

ehelms:
That's an interesting thought; so its a question of treating it as package centric or config centric.

ehelms:
I took a hybrid approach based on my usage building Foreman Proxy SCL over and over. This is updated for that hybrid approach.

obal/data/playbooks/mock/mock.yaml
```diff
@@ -0,0 +1,17 @@
...
+  gather_facts: no
+  roles:
+    - build_srpm
+
+- hosts:
+    - packages
+  serial: 1
+  gather_facts: no
+  tasks:
+    - shell: "mock --recurse --chain -r {{ mock_config }} --localrepo {{ inventory_dir }}/mock_builds SRPMs/*"
```
ekohl: In build_srpm we have a variable for this so we can change it:

https://github.com/theforeman/obal/blob/2f21ce6999b337bb1efd1f8c6cc12287242510d4/obal/data/roles/build_srpm/defaults/main.yaml#L3

Since these are usually the same, it might be nice to also make it a variable here.

Maybe the same goes for `mock_builds`. We should include those in our `.gitignore` and recommend other projects to do the same.

obal/data/roles/mock/tasks/main.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Create SRPM
```
ekohl: Do we want to reuse the `build_srpm` role for this?

ehelms:
Fully updated with tests now

tests/fixtures/mockbin/mock
```diff
@@ -0,0 +1 @@
+mockbin
```
evgeni: meta :)

obal/data/roles/mock/tasks/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Build with mock
+  shell: "mock --recurse --chain -r {{ mock_config }} --localrepo {{ inventory_dir }}/mock_builds SRPMs/*"
```
evgeni: shall `mock_builds` be configurable?

tests/fixtures/testrepo/upstream/mock/el7.cfg
```diff
@@ -0,0 +1,49 @@
+config_opts['root'] = 'foreman-scl-7-x86_64'
```
evgeni: given we're not actually running mock in the tests, would an empty file suffice? or one with a single comment "here shall be a config"?

tests/test_functional.py
```diff
@@ -562,3 +562,14 @@ def test_obal_changelog_custom():
 def test_obal_srpm():
     assert_obal_success(['srpm', 'hello'])
     assert_mockbin_log(['tito build --srpm --offline --output {pwd}/SRPMs'])
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_mock():
+    assert_obal_success(['mock', 'hello', '--config', 'mock/el7.cfg'])
```
evgeni: `--config` is optional, right, so we could drop it from the tests?

evgeni: also, I wonder if we should attempt a test with two packages, that will ensure mock was only called once

obal/data/roles/mock/tasks/main.yaml
ehelms: I suppose it could be if there is a use case for making it configurable.

tests/fixtures/testrepo/upstream/mock/el7.cfg
ehelms: Sure.

ehelms:
@evgeni The move to testing two packages with mock led me down a path of modifying mockbin handling of tito commands and adding a second package to the upstream test repo. I'm going to ignore your approval for now so that you can look over those test changes before I merge.

obal/data/roles/mock/tasks/main.yaml
evgeni: mostly me not liking static stuff :)
maybe pointing this at `~/public_html/myrepo` but then we'd have to drop `{{ inventory_dir }}` too

tests/test_functional.py
```diff
@@ -560,5 +560,19 @@ def test_obal_changelog_custom():
...
+        'tito build --srpm --offline --output {pwd}/SRPMs'
+    ])
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_mock():
+    assert_obal_success(['mock', 'hello', 'foo', '--config', 'mock/el7.cfg'])
+
+    expected_log = [
+        "tito build --srpm --offline --output {pwd}/SRPMs",
```
evgeni: that would need calling twice?

tests/test_functional.py
```diff
@@ -560,5 +560,19 @@ def test_obal_changelog_custom():
...
+    assert_obal_success(['srpm', 'hello', 'foo'])
+    assert_mockbin_log([
+        'tito build --srpm --offline --output {pwd}/SRPMs',
+        'tito build --srpm --offline --output {pwd}/SRPMs'
+    ])
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_mock():
+    assert_obal_success(['mock', 'hello', 'foo', '--config', 'mock/el7.cfg'])
```
evgeni: `--config` is just pointing at the default, can be dropped

evgeni:
@ehelms I am good with your approach, see inline comments for details :)

obal/data/roles/mock/tasks/main.yaml
ehelms: I will make it configurable given that is reasonable to use the public_html location. But I want to keep the default within the project itself to prevent overlap of mock builds across multiple repositories on the same machine (yes, I do this today).

tests/test_functional.py
ehelms: Optional but common to use. Should I write a separate test all together for it?

evgeni: we seldomly test optional params unless they heavily influence the execution logic. so if *I'd* write that test, I'd just write one without `--config` and pretend `--config` works if the main test succeeds. but it also doesn't hurt the way it is now.

obal/data/ansible.cfg
```diff
@@ -2,7 +2,6 @@
 inventory = package_manifest.yaml
 retry_files_enabled = False
 transport = local
-callback_whitelist = profile_tasks
```
ekohl: Could you explain why this is removed?

ehelms: Ohh, I can put that back. It was mostly just muddling my output. I did wonder if anybody actively used this output. If not, its just eating screen space for users.

ekohl: Maybe make it at least a separate commit so it's clearly an unrelated change.

PR #204, opened 2019-09-27
use sys.exit instead of exit

evgeni:
this fixes a pylint error introduced in pylint 2.4.0


evgeni:
@ehelms this "fixes" your issue by making pytest not fail. In reality we need https://github.com/pytest-dev/pytest-xdist/pull/442, but we pin xdist to an older version as anything newer than 1.25 somehow breaks our tests :/

PR #207, opened 2019-09-27
Add ability to update spec files for nodejs and gems

ehelms:
This includes another PR thats required. This adds the ability to update just the spec file of a package (for NPM and Gems) and thus do it en masse. The playbook is a bit messy and I am willing to break it up depending on how folks feel about the design.

### DIFF
--- a/obal/data/modules/changelog.py
+++ b/obal/data/modules/changelog.py
@@ -50,7 +50,7 @@ def main():
 
     changed = False
 
-    if entry and not entry.startswith('-'):
+    if entry is not None and not entry.startswith('-'):
         entry = '- ' + entry
 
     for i, line in enumerate(lines):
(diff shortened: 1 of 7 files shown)


ekohl:
I think the mock commit wasn't intended to be here, correct?

ehelms:
Yes and no. It is not required but while I iterate on any design updates, it's helpful for me to have both the mock and update-spec commands handy locally at the same time :)

ekohl:
Github allows draft PRs for this purpose :)

ehelms:
Truth .. they need to allow setting that after you open it :/

ekohl:
This now needs a rebase.

zjhuntin:
The only thing that this might not do properly is I think that it does doubles up on the changelog instead of printing it once. Everything else, at least for the gems, seems to work properly!

zjhuntin:
My doubling changelog comment is no longer an issue.

PR #208, opened 2019-10-14
rename packages to targets

evgeni:
we're not always working with packages in obal, and in future we might
also work on other things with the same framework

### DIFF
--- a/obal/__init__.py
+++ b/obal/__init__.py
@@ -221,14 +221,14 @@ class ApplicationConfig(object):
         """
         return os.environ.get('OBAL_ANSIBLE_CFG', os.path.join(cls.data_path(), 'ansible.cfg'))
 
-
-def find_playbooks(application_config):
-    """
-    Find all playbooks in the given path.
-    """
-    paths = glob.glob(os.path.join(application_config.playbooks_path(), '*', '*.yaml'))
-    return sorted(Playbook(playbook_path, application_config) for playbook_path in paths if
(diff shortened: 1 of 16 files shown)


obal/__init__.py
```diff
@@ -214,41 +222,41 @@ def ansible_config_path(cls):
         return os.environ.get('OBAL_ANSIBLE_CFG', os.path.join(cls.data_path(), 'ansible.cfg'))
 
 
-def find_playbooks(playbooks_path):
+def find_playbooks(application_config):
```
ekohl: Should this become a method on the `ApplicationConfig`?

tests/test_playbooks.py
```diff
@@ -14,14 +14,14 @@ def playbook_id(fixture_value):
     return fixture_value.name
 
 
-@pytest.fixture(params=obal.find_playbooks(obal.ApplicationConfig.playbooks_path()), ids=playbook_id)
+@pytest.fixture(params=obal.find_playbooks(obal.ApplicationConfig, ids=playbook_id))
```
ekohl:
```suggestion
@pytest.fixture(params=obal.find_playbooks(obal.ApplicationConfig), ids=playbook_id)
```

obal/__init__.py
evgeni: like `ac.find_playbooks`? maybe.

ekohl: Yes. Previously it was a very clean standalone test but now it's coupled to app config anyway. I also think that finding playbooks is also part of the application config.

PR #209, opened 2019-10-22
Fix pylint import not at top level error

ehelms:


evgeni:
hmm, seems not to work as we expected :(

obal/__init__.py
```diff
@@ -237,8 +237,8 @@ def find_targets(inventory_path):
     """
     targets = None
     if os.path.exists(inventory_path):
-        from ansible.inventory.manager import InventoryManager
-        from ansible.parsing.dataloader import DataLoader
+        from ansible.inventory.manager import InventoryManager # pylint: disable=import-outside-toplevel
```
evgeni: great, now this fails on py2, because there the pylint is too old and doesn't know this error.

evgeni:
```suggestion
        from ansible.inventory.manager import InventoryManager # pylint: disable=all
```

evgeni: let's try something

PR #211, opened 2019-10-23
Fix pylint import not at top level error

evgeni:
alternative for #209 (because I can't push to ehelms branch)


evgeni:
@ehelms this is a crude hack, but solves the issue. (and no, you can't disable "bad config" for a line :/)

Issue #210, opened 2019-10-22
Update prerelase version with "obal update"


pcreech:
A good chunk of katello's release steps could be replaced with "obal update", with the exception of setting the prerelease macro.

Perhaps obal can handle this via a "prerelease" argument to obal update?

PR #213, opened 2019-10-28
make obal update adjust the prerelease macro if available

evgeni:
Fixes: #210

### DIFF
--- a/obal/data/roles/fetch_sources/tasks/main.yml
+++ b/obal/data/roles/fetch_sources/tasks/main.yml
@@ -108,6 +108,22 @@
         - '%global mainver'
       when: version is defined
 
+    - name: 'Bump prerelease in specfile'
+      replace:
+        path: "{{ spec_file_path }}"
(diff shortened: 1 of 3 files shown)


obal/data/roles/fetch_sources/tasks/main.yml
```diff
@@ -108,6 +108,22 @@
...
+      with_items:
+        - '%global prerelease'
+      when: prerelease is defined
+
+    - name: 'Remove prerelease in specfile'
+      lineinfile:
+        path: "{{ spec_file_path }}"
+        regexp: '^%global prerelease'
+        state: absent
+      when: prerelease is not defined
```
ehelms: At first, glance, this seems like odd behavior in a `fetch_sources` role.

evgeni: Technically, I agree, but we have all the modifications in that role at the moment.

I can go and try untangle this, but I fear it's a chunk of work.

Issue #214, opened 2019-10-28
fetch_sources role should not modify spec


evgeni:
currently it fiddles around with version/release/prerelease, it shall not

see https://github.com/theforeman/obal/pull/213

PR #216, opened 2019-10-29
split out updating the spec file out of fetch_sources

evgeni:
Fixes: #214


ehelms:
We have two roles right? fetch_sources and setup_sources ? I think that is breeding a lot of my confusion. I recall what the fetch_sources role was born out of a bit more now and I think maybe we need to re-consider it's name? or maybe it's purpose. That is, for something that makes you think it fetches sources, it does updating at the same time. Not your design, but I wondered if you had thoughts from doing this split out of how it could be renamed or broken out such that it makes the responsibility of the role cleaner.

evgeni:
Right. And `setup_sources` is used inside `fetch_sources`, so I can see your confusion.

I'll think about re-naming them in a nicer way.

PR #218, opened 2019-10-30
use the split out obsah library

evgeni:


requirements.txt
```diff
@@ -1 +1,2 @@
 ansible
+git+https://github.com/evgeni/obsah.git@obsah#egg=obsah
```
evgeni: this is obviously temporary until https://github.com/theforeman/obsah/pull/1 is merged and we published `obsah` to pypi.

tests/test_obal.py
```diff
@@ -1,95 +0,0 @@
-import pytest
```
evgeni: this file contained only library tests that now live in `obsah`

requirements.txt
ehelms: It has been merged.

ehelms: Sorry, I meant to ask if the plan was to update this with a released Obsah before merging or to use from git for some amount of development time.

evgeni: I think the released version should be fine. Which is what I've updated the file now for.

requirements.txt
```diff
@@ -1 +1,2 @@
 ansible
+obsah
```
ekohl: Shouldn't this also be in setup.py?

tox.ini
```diff
@@ -8,6 +8,7 @@ envlist = py27,py36,py37,ansible-lint
 
 [testenv]
 deps =
+    -r requirements.txt
```
ekohl: If you include `obsah` in `setup.py` I don't think this is needed.

requirements.txt
evgeni: should, yeah

evgeni:
please don't merge before we don't figure out how to deploy obsah in jenkins

ekohl:
Before we merge this, let's do a final release to tag what's in current master as well?

evgeni:
tagged 0.6.0 before this, and also added obsah to jenkins. let's see what we break.

PR #225, opened 2020-01-14
Fix tag comparison to account for scl packages

ehelms:


pcreech:
ACK

ehelms:
Foreman update that would use this https://github.com/theforeman/foreman-packaging/pull/4583

obal/data/roles/diff_package/tasks/koji.yml
```diff
@@ -5,7 +5,7 @@
     diff_package_changed: False
 
 - name: 'Look up current version of package in {{ diff_package_koji_command }}'
-  shell: "{{ diff_package_koji_command }} list-tagged --quiet --latest {{ item }} {{ inventory_hostname }} \
+  shell: "{{ diff_package_koji_command }} list-tagged --quiet --latest {{ item }} {{ scl + '-' if scl else '' }}{{ inventory_hostname }} \
```
ekohl: Not all SCL packages have the prefix. `foreman` is in SCL but isn't named `tfm-foreman`. The rough guideline is `rubygem-*` and `nodejs-*` now.

ehelms: Fair, and makes this tougher to account for since not everything follows the rules. The addition of this means *most* things will be optimized but not all. The fall back is it will detect a difference and try to build it into Koji (which generally for things like Foreman should happen anyway due to being a nightly package).

ekohl: In my release script I calculate it in the following code: https://gist.github.com/ekohl/d7d01745a8ad9e0532ae092dbf6c22c5#file-verify_packages-L86-L111

ehelms: And you can configure this per package as well to 'disable' it for special ones.

ehelms: We do a similar calculation for the packaging repository part: https://github.com/theforeman/obal/blob/master/obal/data/roles/diff_package/tasks/git_package_info.yml#L3

This code compares what packaging reports to what koji reports to determine if a build is necessary rather than waiting for Koji to report back that a build already exists.

ehelms:
Gonna take this for a test drive given we are doing large rebuilds in foreman-packaging and this should help.

PR #226, opened 2020-02-19
Add a dash at start of changelog message if not present

ehelms:


obal/data/modules/changelog.py
```diff
@@ -50,6 +50,9 @@ def main():
 
     changed = False
 
+    if not changelog.starts_with('-'):
```
ekohl: Looks like a Rubyism
```suggestion
    if not changelog.startswith('-'):
```

evgeni:
Won't this append a dash instead of prepending it?

ekohl:
Sounds like this deserves tests :)

ehelms:
Updated with a test update

PR #227, opened 2020-02-26
Turn off whitelist check since it's not fully functional in it's curr…

pcreech:
…ent state

### DIFF
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -73,6 +73,22 @@ def assert_obal_failure(args):
     run_obal(args, 2)
 
 
+def assert_in_mockbin_log(content):
+    __tracebackhide__ = True
+    expected_log_entry = content.replace('{pwd}', os.getcwd())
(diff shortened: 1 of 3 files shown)


tests/test_functional.py
```diff
@@ -115,7 +115,6 @@ def test_obal_scratch_upstream_hello():
     assert os.path.exists('packages/hello/hello-2.10.tar.gz')
 
     expected_log = [
-        "koji list-pkgs --tag obaltest-nightly-rhel7 --package hello --quiet",
```
ehelms: If one does not exist already, I feel adding a test that configures `build_package_koji_whitelist_check` to true that ensures as part of this command is called would be good since its still an available option.

pcreech: So, I ended up adding two checks, one test for each setting of the variable (true, false), to upstream and downstream sections.

I also added a couple assert functions to assist, ensuring the tests only care if their specific output appears in the log, instead of testing the entire log content.

PR #228, opened 2020-03-03
Don't re-copy files that exist.

pcreech:
Not entirely sure why, but in certain situations, tito will try to recopy assets to the temp directory.  This fails with annex files because they have 444 (r--r--r--) permisisons, and are therefore unwriteable.  Given the narrow window and the very short lifetime of the temp directory (just the life of the tito command), it should be safe to not copy again


zjhuntin:
ACK

pcreech:
Ignoring coveralls, as we don't have test coverage in the tito extension

PR #232, opened 2020-03-19
Add srpm-build to replace tito build

ehelms:
I'm sure this could use some cleaning up, and tests. I would appreciate some early feedback.

### DIFF
--- a/obal/data/module_utils/tito_wrapper.py
+++ b/obal/data/module_utils/tito_wrapper.py
@@ -1,23 +1,9 @@
 """
 A tito wrapper
 """
-import os
-from contextlib import contextmanager
 from subprocess import STDOUT, check_output
 
-
-@contextmanager
-def chdir(directory):
-    """
-    Change the directory in a context manager. Automatically switches back even if an exception
-    occurs.
-    """
-    old = os.getcwd()
-    os.chdir(directory)
-    try:
-        yield
-    finally:
-        os.chdir(old)
+from ansible.module_utils.obal import chdir # pylint:disable=import-error,no-name-in-module
 
 
 def tito(command, directory):
--- a/obal/data/modules/srpm.py
+++ b/obal/data/modules/srpm.py
@@ -1,0 +1,122 @@
+"""
+Build SRPM
+"""
+
+import shutil
+import os
+import subprocess
+from tempfile import mkdtemp
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.obal import chdir # pylint:disable=import-error,no-name-in-module
+
+def run_command(command):
+    """
+    Run a system command
+    """
+    return subprocess.check_output(
+        command,
+        universal_newlines=True,
+        stderr=subprocess.STDOUT
+    )
+
+
+def copy_sources(spec_file, package_dir, sources_dir):
+    """
+    Copy RPM sources to rpmbuild environment
+    """
+    command = ["spectool", "--list-files", spec_file]
+    sources = run_command(command)
+
+    for source in sources.split("\n"):
+        if not source:
+            continue
+
+        source_link = source.split(' ')[1]
+
+        if not source_link.startswith('http'):
+            shutil.copy(os.path.join(package_dir, source_link), sources_dir)
+
+    with chdir(package_dir):
+        run_command(["git-annex", "lock"])
+        annexed_files = run_command(["git-annex", "find", "--include", "*"])
+        annexed_files = annexed_files.splitlines()
+
+        if not annexed_files:
+            return
+
+        run_command(["git-annex", "get"])
+
+        try:
+            run_command(["git-annex", "unlock"])
+        finally:
+            for annex in annexed_files:
+                shutil.copy(annex, sources_dir)
+
+            run_command(["git-annex", "lock", "--force"])
+
+
+def main():
+    """
+    Build a package using tito
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            package=dict(type='str', required=False),
+            scl=dict(type='str', required=False),
+            output=dict(type='path', required=False),
+        )
+    )
+
+    package = module.params['package']
+    output = module.params['output']
+    scl = module.params['scl']
+
+    spec_file = os.path.join(package, '%s.spec' % os.path.basename(package))
+
+    try:
+        base_dir = mkdtemp()
+        sources_dir = os.path.join(base_dir, 'SOURCES')
+        build_dir = os.path.join(base_dir, 'BUILD')
+
+        os.mkdir(sources_dir)
+        os.mkdir(build_dir)
+
+        copy_sources(spec_file, package, sources_dir)
+        shutil.copy(spec_file, base_dir)
+
+        command = ['rpmbuild', '-bs']
+        command += ['--define', '_topdir %s' % base_dir]
+        command += ['--define', '_sourcedir %s' % sources_dir]
+        command += ['--define', '_builddir %s' % build_dir]
+        command += ['--define', '_srcrpmdir %s' % base_dir]
+        command += ['--define', '_rpmdir %s' % base_dir]
+        command += ['--undefine', 'dist']
+
+        if scl:
+            command += ['--define', 'scl %s' %  scl]
+
+        command += [os.path.join(base_dir, '%s.spec' % os.path.basename(package))]
+
+        try:
+            result = run_command(command)
+        except subprocess.CalledProcessError as error:
+            module.fail_json(msg='Failed to srpm build', command=' '.join(command), output=error.output)
+
+        result = result.split("Wrote: ")[-1].rstrip()
+
+        if not os.path.exists(output):
+            os.mkdir(output)
+
+        shutil.copy(result, output)
+        path = os.path.join(output, os.path.basename(result))
+
+        module.exit_json(changed=True, path=path)
+    except subprocess.CalledProcessError as error:
+        module.fail_json(msg='Failed to build srpm', output=error.output)
+    finally:
+        shutil.rmtree(base_dir)
+
+
+if __name__ == '__main__':
+    main()
--- a/obal/data/roles/build_srpm/tasks/main.yaml
+++ b/obal/data/roles/build_srpm/tasks/main.yaml
@@ -1,18 +1,12 @@
 ---
 - name: 'Build SRPM'
-  tito_build:
-    directory: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
-    build_arguments: "{{ build_srpm_tito_build_args }}"
-    srpm: true
-    offline: true
-    dist: "{{ build_srpm_dist }}"
-    scl: "{{ build_srpm_scl }}"
+  srpm:
+    package: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
     output: "{{ build_srpm_output_dir }}"
-    arguments: "{{ build_srpm_tito_args }}"
-    builder: "{{ build_srpm_tito_builder }}"
+    scl: "{{ scl | default(omit) }}"
   register: srpm_build
 
-- name: 'Built SRPM'
+- name: 'Built srpm path'
   debug:
     msg: "Built {{ srpm_build.path }}"
   when: srpm_build is successful
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -657,10 +657,9 @@ def test_obal_changelog_custom():
 @obal_cli_test(repotype='upstream')
 def test_obal_srpm():
     assert_obal_success(['srpm', 'hello', 'foo'])
-    assert_mockbin_log([
-        'tito build --srpm --offline --output {pwd}/SRPMs',
-        'tito build --srpm --offline --output {pwd}/SRPMs'
-    ])
+
+    assert os.path.exists('SRPMs/foo-1.0-1.src.rpm')
+    assert os.path.exists('SRPMs/hello-2.10-2.src.rpm')
 
 
 @obal_cli_test(repotype='upstream')
@@ -668,8 +667,6 @@ def test_obal_mock():
     assert_obal_success(['mock', 'hello', 'foo', '--config', 'mock/el7.cfg'])
 
     expected_log = [
-        "tito build --srpm --offline --output {pwd}/SRPMs",
-        "tito build --srpm --offline --output {pwd}/SRPMs",
-        "mock --recurse --chain -r mock/el7.cfg --localrepo {pwd}/mock_builds SRPMs/foo-1.0-1.src.rpm SRPMs/hello-2.10-1.src.rpm"
+        "mock --recurse --chain -r mock/el7.cfg --localrepo {pwd}/mock_builds SRPMs/foo-1.0-1.src.rpm SRPMs/hello-2.10-2.src.rpm"
     ]
     assert_mockbin_log(expected_log)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
--- a/obal/data/module_utils/obal.py
+++ b/obal/data/module_utils/obal.py
@@ -2,6 +2,8 @@
 Ansible module helper functions for obal modules
 """
 import subprocess
+import os
+from contextlib import contextmanager
 
 
 def specfile_macro_lookup(specfile, macro_str, scl=None, dist=None, macros=None):
@@ -75,3 +77,17 @@ def get_whitelist_status(build_command, tag, package):
     ]
     retcode = subprocess.call(cmd)
     return retcode == 0
+
+
+@contextmanager
+def chdir(directory):
+    """
+    Change the directory in a context manager. Automatically switches back even if an exception
+    occurs.
+    """
+    old = os.getcwd()
+    os.chdir(directory)
+    try:
+        yield
+    finally:
+        os.chdir(old)
--- a/tests/fixtures/testrepo/upstream/packages/foo/foo-source
+++ b/tests/fixtures/testrepo/upstream/packages/foo/foo-source
@@ -1,1 +1,3 @@
 Bar!
+Baz!
+No Foo!
--- a/tests/fixtures/testrepo/upstream/packages/foo/foo.spec
+++ b/tests/fixtures/testrepo/upstream/packages/foo/foo.spec
@@ -19,7 +19,7 @@ A fake RPM for foos and bars.
 %install
 cp . %{buildroot}/
 
-%files foo
+%files
 foo-source
 
 %changelog


obal/data/modules/srpm.py
```diff
@@ -0,0 +1,89 @@
...
+
+    copy_sources(spec_file, package, sources_dir)
+    shutil.copy(spec_file, base_dir)
+
+    command = ['rpmbuild', '-bs']
+    command.append('--define "_topdir %s"' % base_dir)
+    command.append('--define "_sourcedir %s"' % sources_dir)
+    command.append('--define "_builddir %s"' % build_dir)
+    command.append('--define "_srcrpmdir %s"' % base_dir)
+    command.append('--define "_rpmdir %s"' % base_dir)
```
pcreech: From my previous research into this, this doesn't seem to be the _complete_ list of --defines we'll need.

One particular one I can think of is dist, to ensure we generate the correct one for the correct location.

Another is the scl macros.  This _may_ not be too big of a deal with koji's newfound ability to rebuild SRPMS, but i think it's worth noting for completeness that there are a few defines wie'll need.

pcreech: [Link to tito's dist defines](https://github.com/dgoodwin/tito/blob/master/src/tito/builder/main.py#L211-L220)

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,89 @@
...
+    for source in sources.split("\n"):
+        if 'http' not in source:
+            shutil.copy("%s/%s" % (package_dir, source.split(' ')[1]), sources_dir)
+        else:
+            with chdir(package_dir):
+                run_command("git-annex lock")
+                _status, annexed_files = run_command("git-annex find --include='*'")
+                annexed_files = annexed_files.splitlines()
+                run_command("git-annex get")
+                run_command("git-annex unlock")
```
ekohl: I think this should be wrapped in a try-finally

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,89 @@
...
+    return commands.getstatusoutput(command)
+
+def copy_sources(spec_file, package_dir, sources_dir):
+    command = "spectool --list-files %s" % spec_file
+    _status, sources = run_command(command)
+
+    source_files = []
+
+    for source in sources.split("\n"):
+        if 'http' not in source:
```
ekohl: This should be a `startswith()` to avoid issues with local files that have `http` in them (like `fix-http.patch`)

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,89 @@
...
+        argument_spec=dict(
+            package=dict(type='str', required=False),
+            output=dict(type='path', required=False),
+        )
+    )
+
+    package = module.params['package']
+    output = module.params['output']
+
+    spec_file = '%s/%s.spec' % (package, os.path.basename(package))
```
ekohl:
```suggestion
    spec_file = os.path.join(package, '%s.spec' % os.path.basename(package))
```

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,89 @@
+"""
+Build SRPM
+"""
+
+import shutil
+import os
+import commands
```
ekohl: This module is deprecated since Python 2.6 and removed in Python 3 according to https://docs.python.org/2.7/library/commands.html.

obal/data/modules/srpm.py
ekohl: I'd prefer something like:
```suggestion
    defines = {
        "_topdir": base_dir,
        "_sourcedir": sources_dir,
        "_builddir": build_dir,
        "_srcrpmdir": base_dir,
        "_rpmdir": base_dir,
    }
    for key, value in defines.items():
        command.append('--define')
        command.append("%s %s" % (key, value))
```

Then call the command as an array so you don't need to deal with quoting.

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,89 @@
...
+    copy_sources(spec_file, package, sources_dir)
+    shutil.copy(spec_file, base_dir)
+
+    command = ['rpmbuild', '-bs']
+    command.append('--define "_topdir %s"' % base_dir)
+    command.append('--define "_sourcedir %s"' % sources_dir)
+    command.append('--define "_builddir %s"' % build_dir)
+    command.append('--define "_srcrpmdir %s"' % base_dir)
+    command.append('--define "_rpmdir %s"' % base_dir)
+    command.append('%s/%s.spec' % (base_dir, os.path.basename(package)))
```
ekohl:
```suggestion
    command.append(spec_file)
```

obal/data/modules/srpm.py
ehelms: To support dist, I think we'd have to start defining the dists as part of the package manifest metadata and we would have to require that as part of the command requirements. I could add optional support for declaring dist? So that it you can do:

```
obal srpm foo
obal srpm foo --dist=.el7
```

ekohl: Good point. Currently we also set `foremandist` via tito.props.

pcreech: I'd say... We will need to define releasers globally somehow, with specific ability to define the rpm macros we need per releaser that we specify today in tito.props

ehelms: I'm fairly certain the the local dist does not affect the built RPMs dist. As this is controlled by macros inside the buildroot.

obal/data/modules/srpm.py
ehelms: Note this is not the same path as the spec_file from up above. This is within the RPM builddirs.

obal/data/modules/srpm.py
ehelms: I am calling it as an array already. Whats the problem with quoting? I'm open to the idea if there is a bug I am preventing. Otherwise I like the explicitness of the current method.

ekohl: Your current call is `status, result = commands.getstatusoutput(' '.join(command))` but I was suggesting to call it as `subprocess.check_call(command)` (where command is a list). In your version you don't quote the path so if there's a space in `base_dir`, it will fail. Granted, it's unlikely but I've taught myself that doing an array is a way of avoiding a whole class of problems (like using prepared statements instead of trying to quote SQL queries).

ekohl: Btw, if you want to spell it out and still use list, an alternative way of writing this is:
```python
command += ['--define', '_topdir %s' % base_dir]
command += ['--define', '_sourcedir %s' % sources_dir]
# ...
```

ehelms: Ohh right, I can re-test, but calling out to rpmbuild was having issues when I passed the command as an array.

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,111 @@
...
+             )
+    return output
+
+
+def copy_sources(spec_file, package_dir, sources_dir, module):
+    command = ["spectool", "--list-files", spec_file]
+    sources = run_command(command)
+
+    for source in sources.split("\n"):
+        if len(source) == 0:
```
ekohl, line 28:
```suggestion
        if not source:
```

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,111 @@
...
+    shutil.copy(spec_file, base_dir)
+
+    command = ['rpmbuild', '-bs']
+    command += ['--define', '"_topdir %s"' % base_dir]
+    command += ['--define', '"_sourcedir %s"' % sources_dir]
+    command += ['--define', '"_builddir %s"' % build_dir]
+    command += ['--define', '"_srcrpmdir %s"' % base_dir]
+    command += ['--define', '"_rpmdir %s"' % base_dir]
+
+    if scl is not '':
```
ekohl, line 86:
```suggestion
    if scl:
```

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,111 @@
...
+        command += ['--define', '"scl %s"' %  scl]
+
+    command += ['%s/%s.spec' % (base_dir, os.path.basename(package))]
```
ekohl, line 89:
```suggestion
    command += [os.path.join(base_dir, '%s.spec' % os.path.basename(package))]
```

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+    )
+
+    package = module.params['package']
+    output = module.params['output']
+    scl = module.params['scl']
+
+    spec_file = os.path.join(package, '%s.spec' % os.path.basename(package))
+
+    base_dir = mkdtemp()
```
ekohl, line 69: This directory is never cleaned up. Is that intentional?

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,111 @@
...
+
+        if not os.path.exists(output):
+            os.mkdir(output)
+
+        try:
+            shutil.copy(result, output)
+        except Exception as error:
+            module.fail_json(msg='Failed to srpm build', command=' '.join(command), output=error)
+
+        path = "%s/%s" % (output, os.path.basename(result))
```
ekohl, line 103:
```suggestion
        path = os.path.join(output, os.path.basename(result))
```

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,111 @@
...
+    sources = run_command(command)
+
+    for source in sources.split("\n"):
+        if len(source) == 0:
+            continue
+
+        source_link = source.split(' ')[1]
+
+        if not source_link.startswith('http'):
+            shutil.copy("%s/%s" % (package_dir, source_link), sources_dir)
```
ekohl, line 34:
```suggestion
            shutil.copy(os.path.join(package_dir, source_link), sources_dir)
```

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,111 @@
...
+        else:
+            with chdir(package_dir):
+                run_command(["git-annex", "lock"])
+                annexed_files = run_command(["git-annex", "find", "--include='*'"])
```
ekohl, line 38: I think this results in double iteration. You iterate over all sources (in case of bundled NPM packages this may be 50) and now you're iterating over all annexed files which would result in copying every file 50 times, meaning 2500 copies.

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,113 @@
...
+
+        source_link = source.split(' ')[1]
+
+        if not source_link.startswith('http'):
+            shutil.copy(os.path.join(package_dir, source_link), sources_dir)
+
+    with chdir(package_dir):
+        run_command(["git-annex", "lock"])
+        annexed_files = run_command(["git-annex", "find", "--include='*'"])
+        annexed_files = annexed_files.splitlines()
```
ekohl, line 38: Not all packages have annex files. You can probably return early here

ehelms:
@evgeni Can you help me understand this error? Coming from pylint but locally the code works.

```
E:  6, 0: No name 'chdir' in module 'ansible.module_utils' (no-name-in-module)
E:  6, 0: Unable to import 'ansible.module_utils.chdir' (import-error)
```

evgeni:
You can ignore that.
Ansible does weird things for importing from module_utils, and pylint doesn't know about that. So it can't find the code and bails.

ehelms:
@evgeni Curious if you had any thoughts. The error is:

```
TASK [build_srpm : Build SRPM] *************************************************
fatal: [hello]: FAILED! => changed=false 
  msg: Failed to build srpm
  output: |-
    unlock hello-2.10.tar.gz
      this has not yet been committed to git; cannot unlock it
    failed
    git-annex: unlock: 1 failed
```

I do not see this error locally. As I understand it, the test code copies the source in, and does the git annex add and commit before this would be ran: https://github.com/theforeman/obal/blob/master/tests/test_functional.py#L24

ehelms:
@evgeni From your experience, since this passes locally, does this look like an issue with how the repository gets set up, my code or version of git annex?

evgeni:
Most probably the annex version used. Given to this runs on old Ubuntu... You could reconfigure the tests to run on Ubuntu 20.04 (need to explicitly config that, latest still points at 18.04).

ehelms:
That did it, thanks @evgeni !

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,122 @@
...
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.chdir import chdir # pylint:disable=import-error,no-name-in-module
+
+def run_command(command):
+    """
+    Run a system command
+    """
+    return subprocess.check_output(
+        ' '.join(command),
```
evgeni, line 18: the `join` is unnecessary, `subprocess` can handle lists just fine.

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,122 @@
...
+        universal_newlines=True,
+        shell=True,
```
evgeni, line 20: IMHO not needed

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,122 @@
...
+        stderr=subprocess.STDOUT
+    )
+
+
+def copy_sources(spec_file, package_dir, sources_dir):
+    """
+    Copy RPM sources to rpmbuild environment
+    """
+    command = ["spectool", "--list-files", spec_file]
```
evgeni, line 29: not for now, but we have similar calls in `obal/data/playbooks/update/update.yaml` and `obal/data/roles/setup_sources/tasks/specfile.yml`, so it might be useful to extract the listing logic into `module_utils` and use it here and in an own module.

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -1,18 +1,12 @@
...
-    srpm: true
-    offline: true
-    dist: "{{ build_srpm_dist }}"
-    scl: "{{ build_srpm_scl }}"
+  srpm:
+    package: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
     output: "{{ build_srpm_output_dir }}"
-    arguments: "{{ build_srpm_tito_args }}"
-    builder: "{{ build_srpm_tito_builder }}"
+    scl: "{{ scl | default('') }}"
```
evgeni, line 6: `default(omit)`?

obal/data/module_utils/chdir.py
```diff
@@ -0,0 +1,19 @@
+"""
+A chdir context manager
+"""
+import os
+from contextlib import contextmanager
+
+
+@contextmanager
+def chdir(directory):
```
evgeni, line 9: shouldn't we just put that into `obal/data/module_utils/obal.py`? new file for each helper seems a bit too much?

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,122 @@
...
+            run_command(["git-annex", "lock", "--force"])
+
+
+def main():
+    """
+    Build a package using tito
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            package=dict(type='str', required=False),
```
evgeni, line 66: technically that's the path to the package source directory, right?

tests/test_functional.py
```diff
@@ -657,19 +657,16 @@ def test_obal_changelog_custom():
...
 
 @obal_cli_test(repotype='upstream')
 def test_obal_mock():
     assert_obal_success(['mock', 'hello', 'foo', '--config', 'mock/el7.cfg'])
 
     expected_log = [
-        "tito build --srpm --offline --output {pwd}/SRPMs",
-        "tito build --srpm --offline --output {pwd}/SRPMs",
-        "mock --recurse --chain -r mock/el7.cfg --localrepo {pwd}/mock_builds SRPMs/foo-1.0-1.src.rpm SRPMs/hello-2.10-1.src.rpm"
+        "mock --recurse --chain -r mock/el7.cfg --localrepo {pwd}/mock_builds SRPMs/foo-1.0-1.src.rpm SRPMs/hello-2.10-2.src.rpm"
```
evgeni, line 670: why does the revision change from 1 to 2?

or rather: why did it work before!? as the testrepo clearly has -2…

obal/data/modules/srpm.py
ehelms: My experience so far has been that it does not handle them well when combined with calling rpmbuild and specifyin macros that require escaping. I would love it if someone figures out the magic. I spent more time than I should have going round and round with Python subprocess trying to call rpmbuild where macros where defined.

obal/data/modules/srpm.py
ehelms: See https://github.com/theforeman/obal/pull/232#discussion_r468534267

obal/data/roles/build_srpm/tasks/main.yaml
ehelms: That's a thing?! Cool.

obal/data/module_utils/chdir.py
ehelms: Deal.

obal/data/modules/srpm.py
ehelms: Yes it is.

tests/test_functional.py
ehelms: Maybe just how the expectation and the test wrapper were handling things it masked that it didn't matter what was passed? I really don't know lol. The spec in the fixtures has a `2` for it's release.

obal/data/module_utils/chdir.py
ehelms: Next update will remove this file.

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,122 @@
...
+
+        copy_sources(spec_file, package, sources_dir)
+        shutil.copy(spec_file, base_dir)
+
+        command = ['rpmbuild', '-bs']
+        command += ['--define', '"_topdir %s"' % base_dir]
+        command += ['--define', '"_sourcedir %s"' % sources_dir]
+        command += ['--define', '"_builddir %s"' % build_dir]
+        command += ['--define', '"_srcrpmdir %s"' % base_dir]
+        command += ['--define', '"_rpmdir %s"' % base_dir]
```
ekohl, line 94: When you drop the join (https://github.com/theforeman/obal/pull/232/files#r468449779) I think these are double quoted, which is wrong. It's the same as https://github.com/theforeman/puppet-foreman_proxy_content/pull/274#discussion_r454312531
```suggestion
        command += ['--define', '_topdir %s' % base_dir]
        command += ['--define', '_sourcedir %s' % sources_dir]
        command += ['--define', '_builddir %s' % build_dir]
        command += ['--define', '_srcrpmdir %s' % base_dir]
        command += ['--define', '_rpmdir %s' % base_dir]
```

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,122 @@
...
+        command += ['--undefine', 'dist']
+
+        if scl:
+            command += ['--define', '"scl %s"' %  scl]
```
ekohl, line 98: And this one too
```suggestion
            command += ['--define', 'scl %s' %  scl]
```

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,122 @@
...
+
+            if not os.path.exists(output):
+                os.mkdir(output)
+
+            shutil.copy(result, output)
+            path = os.path.join(output, os.path.basename(result))
+
+            module.exit_json(changed=True, path=path)
+        except subprocess.CalledProcessError as error:
+            module.fail_json(msg='Failed to srpm build', command=' '.join(command), output=error.output)
```
ekohl, line 114: Should this be just after `run_command` is called? That's the only thing that can raise a `CalledProcessError`

obal/data/modules/srpm.py
```diff
@@ -0,0 +1,122 @@
...
+            continue
+
+        source_link = source.split(' ')[1]
+
+        if not source_link.startswith('http'):
+            shutil.copy(os.path.join(package_dir, source_link), sources_dir)
+
+    with chdir(package_dir):
+        run_command(["git-annex", "lock"])
+        annexed_files = run_command(["git-annex", "find", "--include='*'"])
```
evgeni, line 43:
```suggestion
        annexed_files = run_command(["git-annex", "find", "--include", "*"])
```

that plus the suggestions from @ekohl make it work with no `join` and no `shell` for me

evgeni: FWIW, this is the patch I have applied locally:

```patch
% git diff
diff --git obal/data/modules/srpm.py obal/data/modules/srpm.py
index 6255fda..e7d8fbf 100644
--- obal/data/modules/srpm.py
+++ obal/data/modules/srpm.py
@@ -15,9 +15,8 @@ def run_command(command):
     Run a system command
     """
     return subprocess.check_output(
-        ' '.join(command),
+        command,
         universal_newlines=True,
-        shell=True,
         stderr=subprocess.STDOUT
     )
 
@@ -40,7 +39,7 @@ def copy_sources(spec_file, package_dir, sources_dir):
 
     with chdir(package_dir):
         run_command(["git-annex", "lock"])
-        annexed_files = run_command(["git-annex", "find", "--include='*'"])
+        annexed_files = run_command(["git-annex", "find", "--include", "*"])
         annexed_files = annexed_files.splitlines()
 
         if not annexed_files:
@@ -87,15 +86,15 @@ def main():
         shutil.copy(spec_file, base_dir)
 
         command = ['rpmbuild', '-bs']
-        command += ['--define', '"_topdir %s"' % base_dir]
-        command += ['--define', '"_sourcedir %s"' % sources_dir]
-        command += ['--define', '"_builddir %s"' % build_dir]
-        command += ['--define', '"_srcrpmdir %s"' % base_dir]
-        command += ['--define', '"_rpmdir %s"' % base_dir]
+        command += ['--define', '_topdir %s' % base_dir]
+        command += ['--define', '_sourcedir %s' % sources_dir]
+        command += ['--define', '_builddir %s' % build_dir]
+        command += ['--define', '_srcrpmdir %s' % base_dir]
+        command += ['--define', '_rpmdir %s' % base_dir]
         command += ['--undefine', 'dist']
 
         if scl:
-            command += ['--define', '"scl %s"' %  scl]
+            command += ['--define', 'scl %s' %  scl]
 
         command += [os.path.join(base_dir, '%s.spec' % os.path.basename(package))]
 
```

obal/data/module_utils/tito_wrapper.py
```diff
@@ -1,23 +1,9 @@
...
-    Change the directory in a context manager. Automatically switches back even if an exception
-    occurs.
-    """
-    old = os.getcwd()
-    os.chdir(directory)
-    try:
-        yield
-    finally:
-        os.chdir(old)
+from ansible.module_utils.chdir import chdir # pylint:disable=import-error,no-name-in-module
```
evgeni, line 6:
```suggestion
from ansible.module_utils.obal import chdir # pylint:disable=import-error,no-name-in-module
```

obal/data/modules/srpm.py
evgeni: #264

PR #233, opened 2020-03-31
Switch to the use of dnf repoclosure

ehelms:


tests/test_functional.py
```diff
@@ -75,10 +75,20 @@ def assert_obal_failure(args):
...
-        log = mockbinlog.read().strip()
-        assert expected_log_entry in log
+
+    if isinstance(content, list):
+        for entry in content:
+            expected_log_entry = entry.replace('{pwd}', os.getcwd())
+
+            with open(os.environ['MOCKBIN_LOG']) as mockbinlog:
+                log = mockbinlog.read().strip()
+                assert expected_log_entry in log
```
ehelms: Add the ability here to pass an array of statements to be checked for existence within the log. This is because the new repoclosure uses a temporary directory that is random on each run which makes it impossible to check against a traditional string comparison.

ekohl: Wouldn't it make more sense to do the entry comparison loop inside the with here? Now you're opening and reading the file for every entry while `mockbinlog` remains unchanged. You can even move the reading to before the entire if/else.

ehelms: Sure, that could make sense as a performance improvement.

ehelms: Makes sense as an improvement. Here ya go https://github.com/theforeman/obal/pull/234

PR #234, opened 2020-04-01
Run assert loop inside already open file in tests

ehelms:


tests/test_functional.py
```diff
@@ -77,11 +77,11 @@ def assert_in_mockbin_log(content):
     __tracebackhide__ = True
 
     if isinstance(content, list):
-        for entry in content:
-            expected_log_entry = entry.replace('{pwd}', os.getcwd())
+        with open(os.environ['MOCKBIN_LOG']) as mockbinlog:
```
ekohl, line 80: You can actually move this to before the if/else and drop it in the else branch as well. Then you reduce the duplication even further. Since the file is read and stored in a variable, you don't need to do the assertion within the `with` block.

ekohl:
Thanks! Much cleaner

PR #235, opened 2020-04-01
Add dnf and dnf-plugins-core to Dockerfile

ehelms:


Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM centos:7
 RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
     yum -y install epel-release git && \
     yum -y install python2-pip && \
+    yum -y install dnf-plugins-core dnf && \
```
ekohl, line 6: Any particular reason we don't make this a single `yum -y install` line? Does it make sense to do it in https://github.com/theforeman/obal/blob/master/obal/data/playbooks/setup/setup.yaml?

ehelms: From a container perspective this is all one layer, doing each line allows some grouping of common items. Doesn't hurt anything. I can add it to setup as well.

ehelms:
The one reason I kept it as a centos:7 container for now was to have an environment that mimics our Jenkins nodes. That was incredibly useful when debugging the use of `dnf repoclosure`.

Dockerfile
ekohl: IMHO we should either have all runtime deps in setup or everything in the dockerfile. The dockerfile calls setup anyway. I'd also be fine with installing all deps here and not calling setup. It would be more efficient in caching layers.

ehelms:
Ahh, duh on my part. I get your point now. Setup is updated only now.

sthirugn:
ACK

ehelms:
Passing https://travis-ci.org/github/theforeman/obal/builds/670112180?utm_source=github_status&utm_medium=notification

PR #237, opened 2020-04-03
Allow inventory as a directory by using project_dir from Obsah

ehelms:


ehelms:
Requires https://github.com/theforeman/obsah/pull/3

The idea is to allow breaking up big package manifests into smaller files inside a directory.

obal/__init__.py
```diff
@@ -60,7 +60,12 @@ def inventory_path():
         """
         Return the inventory path
         """
-        return os.environ.get('OBAL_INVENTORY', os.path.join(os.getcwd(), 'package_manifest.yaml'))
+        inventory = os.path.join(os.getcwd(), 'package_manifest.yaml')
+
+        if os.path.isdir('package_manifests'):
+            inventory = os.path.join(os.getcwd(), 'package_manifests')
+
+        return os.environ.get('OBAL_INVENTORY', inventory)
```
ekohl, line 68: Would this be clearer?

```suggestion
        try:
            return os.environ['OBAL_INVENTORY']
        except KeyError:
            if os.path.isdir('package_manifests'):
                return os.path.join(os.getcwd(), 'package_manifests')
            else:
                return os.path.join(os.getcwd(), 'package_manifest.yaml')
```

PR #240, opened 2020-04-07
Introduce koji_tags variable

ehelms:


obal/data/roles/diff_package/tasks/koji.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- set_fact:
+    koji_tags: "{{ diff_package_tags }}"
+  when: diff_package_tags is defined
```
ehelms, line 4: Keeps backwards compatibility. I debated whether we should lock our older branches to a version of Obal instead and keep the code clean going forward.

ehelms:
Alllows the transition to dropping tito to be more step by step than big bang.

PR #230, opened 2020-03-16
Use koji directly for release tasks

ehelms:
The idea here is to move away from relying on tito releasers and whitelists and using package_manifest and koji directly for build tasks. Given this requires transition within a package_manifest I am attempting to build it such that both functionality is supported for some period of time based on available configuration in the package_manifest.

PR #242, opened 2020-04-09
Add modules for nevr and checking if build exists in koji

ehelms:


ehelms:
See example https://github.com/theforeman/foreman-packaging/pull/4989

ehelms:
Now that I built these I wondered if this still made sense as a stand-alone role or if this should just be part of the koji build package role or the koji_build module introduced here (https://github.com/theforeman/obal/pull/230/files#diff-de47576edf20c669c22f90614e98fc65) to make that module more idempotent directly.

obal/data/modules/check_koji_build.py
```diff
@@ -0,0 +1,39 @@
+"""
+Check if build exists in Koji
+"""
+
+import os
+from subprocess import STDOUT, check_output
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.obal import get_specfile_nevr
```
evgeni, line 9: this looks unused

obal/data/modules/check_koji_build.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+    nevr = module.params['nevr']
+    tag = module.params['tag']
+    package = module.params['package']
+
+    command = ['koji', 'latest-build', '--quiet', tag, package]
+
+    try:
+        build = check_output(command, stderr=STDOUT, universal_newlines=True)
```
evgeni, line 30: this could use the `koji` wrapper instead?

obal/data/modules/check_koji_build.py
```diff
@@ -0,0 +1,39 @@
...
+        build = build.split(' ')[0]
+
+        module.exit_json(changed=False, tagged_version=build, exists=(nevr == build))
```
evgeni, line 33: the build contains `%dist`, while `nevr` does not (as we do `--undefine dist` when querying it), so this will never be true.

obal/data/roles/diff_package/tasks/koji.yml
```diff
@@ -1,34 +1,5 @@
...
-  no_log: True
-
-- debug:
-    msg:
-      - "Git version: {{ git_package_version }}"
-      - "Tagged version: {{ item }}"
-      - "Changed: {{ diff_package_changed }}"
-  with_items: "{{ koji_package_versions }}"
+- include_tasks: koji_tag.yml
+  loop: "{{ koji_tags }}"
```
evgeni, line 3: for this to work, `koji_tags` needs to be a list, but it is a dict. do you want `with_dict`?

evgeni: or you need to convert it on the fly: https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#with-dict

obal/data/roles/diff_package/tasks/koji_tag.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Find local NEVR for package
+  nevr:
+    spec_file: "{{ spec_file_path }}"
+    scl: "{{ tag.scl | default(None) }}"
+    dist: "{{ tag.dist | default(None) }}"
```
evgeni, line 6: shouldn't these be `default(omit)`? `None` is so pythonic, but this is YAML/Ansible :)

obal/data/modules/check_koji_build.py
evgeni: or do you plan to pass `dist` from the tag definition in the manifest, and it's just not in the examples below?

obal/data/modules/nevr.py
```diff
@@ -0,0 +1,38 @@
+"""
+Calculate NEVRA for a package
```
evgeni, line 2:
```suggestion
Calculate NEVR for a package
```

obal/data/modules/nevr.py
```diff
@@ -0,0 +1,38 @@
...
+
+def main():
+    """
+    Calculate name-epoch-version-release for a specfile
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            spec_file=dict(type='str', required=True),
+            scl=dict(type='str', required=False, default=None),
+            dist=dict(type='str', required=False, default=None),
```
evgeni, line 17: required=False will result in `None` in the case of no input, no need to set `default`

obal/data/roles/diff_package/tasks/koji.yml
```diff
@@ -1,34 +1,5 @@
...
-  changed_when: False
-
-- name: "Set koji_package_versions"
-  set_fact:
-    koji_package_versions: "{{ koji_package_versions.results | map(attribute='stdout') | list }}"
-  no_log: True
-
-- name: "Calculate diff_package_changed"
-  set_fact:
-    diff_package_changed: "{{ not item.startswith(git_package_version) or diff_package_changed }}"
```
evgeni, line 25: nothing in the new code sets `diff_package_changed`, so the role will not have a "result" to act on?

evgeni:
> Now that I built these I wondered if this still made sense as a stand-alone role or if this should just be part of the koji build package role or the koji_build module introduced here (https://github.com/theforeman/obal/pull/230/files#diff-de47576edf20c669c22f90614e98fc65) to make that module more idempotent directly.

We have a `check` command, that also uses this role, but w/o the "do the build" step. Essentially to verify that git and koji are in sync. I've not used it in a while as the old diff_package was semi-broken, but I think it's valuable to have. What if we move the actual checking code into a helper and then make both `check_koji_build` and `koji_build` use it, one reporting the state, the other doing the build? Or maybe, we could just teach `koji_build` check_mode and then the check playbook would force that? Yeah, that's probably the cleanest solution.

ekohl:
This suggests we don't need release bumps just to start building on EL8 anymore. Just run obal release on all those packages. Correct?

obal/data/modules/check_koji_build.py
ehelms: Ahh yes, sorry about that. I forgot to update the fixtures with this push to represent the proposed structure. Next push will include that update.

obal/data/roles/diff_package/tasks/koji.yml
ehelms: See https://github.com/theforeman/obal/pull/242#discussion_r406161701

ehelms:
@evgeni Passing tests and covering the pass use case including backwards compatibility. I decided to just keep diff_package working as expected to not overload this change with the additional element of making koji_build more idempotent. I think this provides a good overall change and sets the stage to keep re-factoring.

obal/data/module_utils/koji_wrapper.py
```diff
@@ -0,0 +1,19 @@
...
+
+class KojiCommandError(Exception):
+    """Raised when Koji command fails"""
+
+def koji(command, executable):
+    """
+    Run a koji command
+    """
+    if not executable:
+        executable = 'koji'
```
evgeni, line 14:
```suggestion
def koji(command, executable=None):
    """
    Run a koji command
    """
    if executable is None:
        executable = 'koji'
```

Python does not allow Ruby syntax ;)

tests/fixtures/mockbin/mockbin
```diff
@@ -77,29 +77,36 @@ if prog == 'tito':
 elif prog in ['koji', 'brew']:
     parser = argparse.ArgumentParser()
     subparsers = parser.add_subparsers()
+
     parser_listpkgs = subparsers.add_parser('list-pkgs')
     parser_listpkgs.add_argument('--tag')
     parser_listpkgs.add_argument('--package')
     parser_listpkgs.add_argument('--quiet', action='store_true')
-    parser_listtagged = subparsers.add_parser('list-tagged')
```
evgeni, line 84: This is still used in `koji_old.yml`, but we never call that in our tests now. It feels like we'll end up with bit-rot if we drop the tests?

obal/data/modules/nevr.py
```diff
@@ -0,0 +1,38 @@
+"""
+Calculate name-epoch:version-release for a package
```
evgeni, line 2: Something tells me, this should be called `rpm_nevr` ;)

tests/fixtures/mockbin/mockbin
ehelms: I added it back and switch the downstream test repo to use the old syntax. So that there is now a mix of the two throughout the tests until we drop diff_package_tags

obal/data/module_utils/obal.py
```diff
@@ -4,21 +4,30 @@
...
-    cmd = [
+    command = [
         'rpmspec',
         '--query',
         '--queryformat',
         macro_str,
-        '--undefine',
-        'dist',
-        '--undefine',
-        'foremandist',
```
evgeni, line 17: Is this intentional? When you execute this on a machine which has `foreman-build` installed, this will get weird NEVRs.

ehelms: Yes, my intent was to have the projects define this if need be using the new ability to pass macros per tag since this was quite Foreman specific.

evgeni: ok

ehelms:
@evgeni thanks! This supports backwards compatability with diff_package_tags but not the recently updated `koji_tags` as a per list change. Therefore, to not break things https://github.com/theforeman/foreman-packaging/pull/4989 has to go in at the same time if you could review it please.

PR #243, opened 2020-04-13
Add modules and command to verify a tag against local content

ehelms:
Allows checking if what is local is in the tag on Koji and vice versa.

```
obal verify-tag --tag foreman-nightly-rhel7
obal verify-tag --all
```

@ekohl This would implement some of the gaps you requested

### DIFF
--- a/obal/data/modules/koji_find_tags.py
+++ b/obal/data/modules/koji_find_tags.py
@@ -1,0 +1,29 @@
+"""
+Find all defined Koji tags in inventory
+"""
+
+from ansible.module_utils.basic import AnsibleModule
+
+def main():
+    """
+    Find all defined Koji tags in inventory
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            packages=dict(type='dict', required=True)
+        )
+    )
+
+    packages = module.params['packages']
+
+    tags = set()
+
+    for attributes in packages.values():
+        if 'koji_tags' in attributes:
+            tags.update(tag['name'] for tag in attributes['koji_tags'])
+
+    module.exit_json(changed=False, tags=sorted(tags))
+
+
+if __name__ == '__main__':
+    main()
--- a/obal/data/modules/koji_verify_tag.py
+++ b/obal/data/modules/koji_verify_tag.py
@@ -1,0 +1,68 @@
+"""
+Verify packages against a tag in Koji
+"""
+
+import os
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.obal import get_specfile_name # pylint:disable=import-error,no-name-in-module
+from ansible.module_utils.koji_wrapper import koji # pylint:disable=import-error,no-name-in-module
+
+def main():
+    """
+    Verify packages against a tag in Koji
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            packages=dict(type='dict', required=True),
+            tag=dict(type='str', required=True),
+            directory=dict(type='str', required=True),
+            koji_executable=dict(type='str', require=False, default='koji')
+        )
+    )
+
+    packages_for_tag = set()
+
+    for (package, attributes) in module.params['packages'].items():
+        tag = None
+
+        if 'koji_tags' in attributes:
+            for koji_tag in attributes['koji_tags']:
+                if module.params['tag'] == koji_tag['name']:
+                    tag = koji_tag
+
+            if tag and 'package_base_dir' in attributes:
+                specfile = os.path.join(attributes['package_base_dir'], package, "{}.spec".format(package))
+                scl = tag.get('scl')
+                name = get_specfile_name(os.path.join(module.params['directory'], specfile), scl)
+                packages_for_tag.add(name)
+
+    try:
+        command = ['list-pkgs', '--quiet', '--tag', module.params['tag']]
+        koji_output = koji(
+            command,
+            executable=module.params['koji_executable']
+        )
+
+        packages_in_koji = {item.split(' ', 1)[0] for item in koji_output.split("\n") if item}
+
+    except subprocess.CalledProcessError as error:
+        module.fail_json(changed=False, msg=error.output)
+
+    missing_in_koji = packages_for_tag - packages_in_koji
+    missing_in_git = packages_in_koji - packages_for_tag
+
+    if missing_in_koji or missing_in_git:
+        module.fail_json(
+            changed=True,
+            msg="Package differences found",
+            missing_in_git=sorted(missing_in_git),
+            missing_in_koji=sorted(missing_in_koji)
+        )
+    else:
+        module.exit_json(changed=False)
+
+
+if __name__ == '__main__':
+    main()
--- a/tests/fixtures/testrepo/upstream/package_manifest.yaml
+++ b/tests/fixtures/testrepo/upstream/package_manifest.yaml
@@ -10,6 +10,8 @@ packages:
     koji_tags:
       - name: obaltest-nightly-rhel7
         dist: '.el7'
+      - name: obaltest-nightly-el8
+        dist: '.el8'
   hosts:
     hello:
       nightly_package_tito_releaser_args:
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -129,6 +129,8 @@ def test_obal_check_upstream_hello():
     expected_log = [
         "koji buildinfo hello-2.10-2.el7",
         "koji latest-build --quiet obaltest-nightly-rhel7 hello",
+        "koji buildinfo hello-2.10-2.el8",
+        "koji latest-build --quiet obaltest-nightly-el8 hello"
     ]
     assert_mockbin_log(expected_log)
 
@@ -169,6 +171,9 @@ def test_obal_scratch_with_koji_upstream_hello():
         "koji build obaltest-nightly-rhel7 /tmp/SRPMs/hello-2.10-2.src.rpm --scratch",
         "koji watch-task 1234",
         "koji taskinfo -v 1234",
+        "koji build obaltest-nightly-el8 /tmp/SRPMs/hello-2.10-2.src.rpm --scratch",
+        "koji watch-task 1234",
+        "koji taskinfo -v 1234",
     ]
     assert_mockbin_log(expected_log)
     assert os.path.exists('/tmp/SRPMs/hello-2.10-2.src.rpm')
@@ -181,7 +186,8 @@ def test_obal_scratch_with_koji_upstream_hello_nowait():
     assert os.path.exists('packages/hello/hello-2.10.tar.gz')
 
     expected_log = [
-        "koji build obaltest-nightly-rhel7 /tmp/SRPMs/hello-2.10-2.src.rpm --scratch"
+        "koji build obaltest-nightly-rhel7 /tmp/SRPMs/hello-2.10-2.src.rpm --scratch",
+        "koji build obaltest-nightly-el8 /tmp/SRPMs/hello-2.10-2.src.rpm --scratch"
     ]
     assert_mockbin_log(expected_log)
     assert os.path.exists('/tmp/SRPMs/hello-2.10-2.src.rpm')
@@ -197,6 +203,12 @@ def test_obal_release_with_koji_upstream_hello():
         "koji latest-build --quiet obaltest-nightly-rhel7 hello",
         "koji build obaltest-nightly-rhel7 /tmp/SRPMs/hello-2.10-2.src.rpm",
         "koji watch-task 1234",
+        "koji taskinfo -v 1234",
+        "koji buildinfo hello-2.10-2.el8",
+        "koji latest-build --quiet obaltest-nightly-el8 hello",
+        "koji latest-build --quiet obaltest-nightly-el8 hello",
+        "koji build obaltest-nightly-el8 /tmp/SRPMs/hello-2.10-2.src.rpm",
+        "koji watch-task 1234",
         "koji taskinfo -v 1234"
     ]
     assert_mockbin_log(expected_log)
@@ -209,7 +221,13 @@ def test_obal_release_with_koji_upstream_existing_build():
     expected_log = [
         "koji buildinfo package-with-existing-build-1.0-1.el7",
         "koji latest-build --quiet obaltest-nightly-rhel7 package-with-existing-build",
-        "koji tag-build obaltest-nightly-rhel7 package-with-existing-build-1.0-1.el7"
+        "koji tag-build obaltest-nightly-rhel7 package-with-existing-build-1.0-1.el7",
+        "koji buildinfo package-with-existing-build-1.0-1.el8",
+        "koji latest-build --quiet obaltest-nightly-el8 package-with-existing-build",
+        "koji latest-build --quiet obaltest-nightly-el8 package-with-existing-build",
+        "koji build obaltest-nightly-el8 /tmp/SRPMs/package-with-existing-build-1.0-1.src.rpm",
+        "koji watch-task 1234",
+        "koji taskinfo -v 1234"
     ]
     assert_mockbin_log(expected_log)
 
@@ -221,7 +239,14 @@ def test_obal_release_with_koji_upstream_whitelist_check():
     expected_log = [
         "koji buildinfo package-with-existing-build-1.0-1.el7",
         "koji latest-build --quiet obaltest-nightly-rhel7 package-with-existing-build",
-        "koji tag-build obaltest-nightly-rhel7 package-with-existing-build-1.0-1.el7"
+        "koji tag-build obaltest-nightly-rhel7 package-with-existing-build-1.0-1.el7",
+        "koji buildinfo package-with-existing-build-1.0-1.el8",
+        "koji latest-build --quiet obaltest-nightly-el8 package-with-existing-build",
+        "koji list-pkgs --tag obaltest-nightly-el8 --package package-with-existing-build --quiet",
+        "koji latest-build --quiet obaltest-nightly-el8 package-with-existing-build",
+        "koji build obaltest-nightly-el8 /tmp/SRPMs/package-with-existing-build-1.0-1.src.rpm",
+        "koji watch-task 1234",
+        "koji taskinfo -v 1234"
     ]
     assert_mockbin_log(expected_log)
 
@@ -235,6 +260,8 @@ def test_obal_release_upstream_hello():
     expected_log = [
         "koji buildinfo hello-2.10-2.el7",
         "koji latest-build --quiet obaltest-nightly-rhel7 hello",
+        "koji buildinfo hello-2.10-2.el8",
+        "koji latest-build --quiet obaltest-nightly-el8 hello",
         "tito release --yes dist-git",
         "koji watch-task 1234",
         "koji taskinfo -v 1234",
@@ -267,7 +294,9 @@ def test_obal_release_upstream_hello_nowait():
     expected_log = [
         "koji buildinfo hello-2.10-2.el7",
         "koji latest-build --quiet obaltest-nightly-rhel7 hello",
-        "tito release --yes dist-git"
+        "koji buildinfo hello-2.10-2.el8",
+        "koji latest-build --quiet obaltest-nightly-el8 hello",
+        "tito release --yes dist-git",
     ]
     assert_mockbin_log(expected_log)
 
@@ -281,6 +310,8 @@ def test_obal_release_upstream_hello_waitrepo():
     expected_log = [
         "koji buildinfo hello-2.10-2.el7",
         "koji latest-build --quiet obaltest-nightly-rhel7 hello",
+        "koji buildinfo hello-2.10-2.el8",
+        "koji latest-build --quiet obaltest-nightly-el8 hello",
         "tito release --yes dist-git",
         "koji watch-task 1234",
         "koji taskinfo -v 1234",
@@ -732,5 +763,24 @@ def test_obal_mock():
 
     expected_log = [
         "mock --recurse --chain -r mock/el7.cfg --localrepo {pwd}/mock_builds SRPMs/foo-1.0-1.src.rpm SRPMs/hello-2.10-2.src.rpm"
+    ]
+    assert_mockbin_log(expected_log)
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_verify_tag():
+    assert_obal_success(['verify-koji-tag', 'obaltest-nightly-rhel7'])
+
+    expected_log = ['koji list-pkgs --quiet --tag obaltest-nightly-rhel7']
+    assert_mockbin_log(expected_log)
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_verify_tag_all():
+    assert_obal_success(['verify-koji-tag', 'all'])
+
+    expected_log = [
+        'koji list-pkgs --quiet --tag obaltest-nightly-el8',
+        'koji list-pkgs --quiet --tag obaltest-nightly-rhel7'
     ]
     assert_mockbin_log(expected_log)
--- a/obal/data/playbooks/verify-koji-tag/metadata.obal.yaml
+++ b/obal/data/playbooks/verify-koji-tag/metadata.obal.yaml
@@ -1,0 +1,5 @@
+help: Verify packages against a Koji tag
+variables:
+  tag:
+    parameter: tag
+    help: Koji tag to verify against
--- a/obal/data/playbooks/verify-koji-tag/verify-koji-tag.yaml
+++ b/obal/data/playbooks/verify-koji-tag/verify-koji-tag.yaml
@@ -1,0 +1,6 @@
+---
+- hosts:
+    - localhost
+  gather_facts: false
+  roles:
+    - verify_koji_tag
--- a/obal/data/roles/verify_koji_tag/defaults/main.yml
+++ b/obal/data/roles/verify_koji_tag/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+verify_koji_tag_all: false
--- a/obal/data/roles/verify_koji_tag/tasks/all.yaml
+++ b/obal/data/roles/verify_koji_tag/tasks/all.yaml
@@ -1,0 +1,10 @@
+---
+- name: Find all tags
+  koji_find_tags:
+    packages: "{{ hostvars }}"
+  register: tags
+
+- include_tasks: "verify.yaml"
+  loop: "{{ tags.tags }}"
+  loop_control:
+    loop_var: tag_name
--- a/obal/data/roles/verify_koji_tag/tasks/main.yaml
+++ b/obal/data/roles/verify_koji_tag/tasks/main.yaml
@@ -1,0 +1,8 @@
+---
+- when: tag != 'all'
+  include_tasks: "verify.yaml"
+  vars:
+    tag_name: "{{ tag }}"
+
+- when: tag == 'all'
+  include_tasks: "all.yaml"
--- a/obal/data/roles/verify_koji_tag/tasks/verify.yaml
+++ b/obal/data/roles/verify_koji_tag/tasks/verify.yaml
@@ -1,0 +1,22 @@
+---
+- name: Verify tag
+  koji_verify_tag:
+    tag: "{{ tag_name }}"
+    packages: "{{ hostvars }}"
+    directory: "{{ lookup('env', 'PWD') }}"
+    koji_executable: "{{ koji_executable|default('koji') }}"
+  register: missing
+
+- name: "Packages in Koji but not in git for {{ tag_name }}"
+  debug:
+    msg: "{{ missing.missing_in_git }}"
+  when:
+    - missing is changed
+    - missing.missing_in_git|length > 0
+
+- name: "Packages missing in Koji for {{ tag_name }}"
+  debug:
+    msg: "{{ missing.missing_in_koji }}"
+  when:
+    - missing is changed
+    - missing.missing_in_koji|length > 0
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
 
     packages=find_packages(exclude=['contrib', 'docs', 'tests']),
 
-    install_requires=['ansible >= 2.5', 'obsah'],
+    install_requires=['ansible >= 2.5', 'obsah >= 0.0.2'],
 
     extras_require={
         'argcomplete': ['argcomplete'],
--- a/tests/fixtures/help/verify-koji-tag.txt
+++ b/tests/fixtures/help/verify-koji-tag.txt
@@ -1,0 +1,15 @@
+usage: obal verify-koji-tag [-h] [-v] [-e EXTRA_VARS] tag
+
+Verify packages against a Koji tag
+
+positional arguments:
+  tag                   Koji tag to verify against
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -v, --verbose         verbose output
+
+advanced arguments:
+  -e EXTRA_VARS, --extra-vars EXTRA_VARS
+                        set additional variables as key=value or YAML/JSON, if
+                        filename prepend with @
--- a/tests/test_playbooks.py
+++ b/tests/test_playbooks.py
@@ -20,7 +20,7 @@ def playbook(request):
 
 
 def test_takes_target_argument(playbook):
-    expected = playbook.name not in ('setup', 'cleanup-copr')
+    expected = playbook.name not in ('setup', 'cleanup-copr', 'verify-koji-tag')
     assert playbook.takes_target_parameter == expected
 
 
--- a/tox.ini
+++ b/tox.ini
@@ -20,7 +20,7 @@ deps =
 commands =
     check-manifest --ignore tox.ini,.ansible-lint,tests*,tests/**,docs*,docs/**,*/**/*.pyc,*/**/__pycache__,.yamllint
     python setup.py check -m -s
-    pytest -v -n 4 --flakes --pylint --pylint-rcfile={toxinidir}/.pylintrc --cov=obal --cov-report term --cov-report xml --boxed {posargs}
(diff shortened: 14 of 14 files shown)


obal/data/modules/verify_tag.py
```diff
@@ -0,0 +1,65 @@
...
+                scl = tag['scl'] if 'scl' in tag.keys() else None
+                package = get_specfile_name(os.path.join(directory, specfile), scl)
+                packages_for_tag.append(package)
+
+    try:
+        command = ['koji', 'list-pkgs', '--quiet', '--tag', tag_name]
+        packages_in_koji = check_output(command, stderr=STDOUT, universal_newlines=True)
+        packages_in_koji = [item.split(' ')[0] for item in packages_in_koji.split("\n")]
+        packages_in_koji = [package for package in packages_in_koji if len(package) > 0]
+    except subprocess.CalledProcessError as error:
```
evgeni, line 47: that's not imported and thus cannot be caught.

obal/data/modules/verify_tag.py
```diff
@@ -0,0 +1,65 @@
+"""
+Verify packages against a tag in Koji
+"""
+
+import os
+from subprocess import STDOUT, check_output
```
evgeni, line 6:
```suggestion
from subprocess import STDOUT, check_output, CalledProcessError
```

obal/data/modules/verify_tag.py
evgeni, line 47:
```suggestion
    except CalledProcessError as error:
```

obal/data/roles/verify_tag/tasks/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- include_tasks: "verify.yaml"
+  when: verify_tag_all == False
```
evgeni, line 3: `[E601] Don't compare to literal True/False` ;)

obal/data/roles/verify_tag/tasks/main.yaml
```diff
@@ -0,0 +1,6 @@
...
+
+- include_tasks: "all.yaml"
+  when: verify_tag_all == True
```
evgeni, line 6: `[E601] Don't compare to literal True/False` ;)

evgeni:
Should the modules (not the command/playbook) be prefixed with `koji_` or so?

tests/test_functional.py
```diff
@@ -670,3 +670,10 @@ def test_obal_mock():
         "mock --recurse --chain -r mock/el7.cfg --localrepo {pwd}/mock_builds SRPMs/foo-1.0-1.src.rpm SRPMs/hello-2.10-2.src.rpm"
     ]
     assert_mockbin_log(expected_log)
+
+@obal_cli_test(repotype='upstream')
+def test_obal_verify_tag():
+    assert_obal_success(['verify-tag', '--tag', 'obaltest-nightly-rhel7'])
+
+    expected_log = []
```
evgeni, line 678: that's not true ;)

tests/fixtures/testrepo/upstream/package_manifest.yaml
```diff
@@ -10,6 +10,8 @@ packages:
     koji_tags:
       - name: obaltest-nightly-rhel7
         dist: '.el7'
+      - name: obaltest-nightly-el8
+        dist: '.el8'
```
evgeni, line 14: this made some other tests fail as they have not been updated to check for both el7 and el8 things:


```
FAILED tests/test_functional.py::test_obal_check_upstream_hello
FAILED tests/test_functional.py::test_obal_release_upstream_hello
FAILED tests/test_functional.py::test_obal_release_upstream_hello_nowait
FAILED tests/test_functional.py::test_obal_release_upstream_hello_nowaitrepo
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,70 @@
...
+
+            if tag and 'package_base_dir' in attributes:
+                specfile = "%s/%s/%s.spec" % (attributes['package_base_dir'], package, package)
+                scl = tag['scl'] if 'scl' in tag.keys() else None
+                name = get_specfile_name(os.path.join(module.params['directory'], specfile), scl)
+                packages_for_tag.append(name)
+
+    try:
+        command = ['koji', 'list-pkgs', '--quiet', '--tag', tag_name]
+        packages_in_koji = subprocess.check_output(
```
evgeni, line 43: any reason not to use `koji_wrapper.koji` that we have?

evgeni:
nitpicks and test-failures, but overall :+1:

obal/data/modules/koji_find_tags.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            packages=dict(type='dict', required=True)
+        )
+    )
+
+    packages = module.params['packages']
+
+    tags = []
```
ekohl, line 19: I'd suggest to make this a `set()`
```suggestion
    tags = set()
```

obal/data/modules/koji_find_tags.py
```diff
@@ -0,0 +1,31 @@
...
+
+    for (_package, attributes) in packages.items():
+        if 'koji_tags' in attributes:
+            for tag in attributes['koji_tags']:
+                if tag['name'] not in tags:
+                    tags.append(tag['name'])
```
ekohl, line 25:
```suggestion
            tags += set(attributes['koji_tags'])
```

obal/data/modules/koji_find_tags.py
```diff
@@ -0,0 +1,31 @@
...
+        argument_spec=dict(
+            packages=dict(type='dict', required=True)
+        )
+    )
+
+    packages = module.params['packages']
+
+    tags = []
+
+    for (_package, attributes) in packages.items():
```
ekohl, line 21:
```suggestion
    for attributes in packages.values():
```

obal/data/modules/koji_find_tags.py
```diff
@@ -0,0 +1,31 @@
...
+        if 'koji_tags' in attributes:
+            for tag in attributes['koji_tags']:
+                if tag['name'] not in tags:
+                    tags.append(tag['name'])
+
+    module.exit_json(changed=False, tags=tags)
```
ekohl, line 27:
```suggestion
    module.exit_json(changed=False, tags=sorted(tags))
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,70 @@
...
+    missing_in_koji = []
+    missing_in_git = []
+
+    for package in packages_for_tag:
+        if package not in packages_in_koji:
+            missing_in_koji.append(package)
+
+    for package in packages_in_koji:
+        if package not in packages_for_tag:
+            missing_in_git.append(package)
```
ekohl, line 64: Sets are your friend
```suggestion
    missing_in_koji = packages_for_tag - packages_in_koji
    missing_in_git = packages_in_koji - packages_for_tag
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,70 @@
...
+        argument_spec=dict(
+            packages=dict(type='dict', required=True),
+            tag=dict(type='str', required=True),
+            directory=dict(type='str', required=True)
+        )
+    )
+
+    tag_name = module.params['tag']
+
+    packages_for_tag = []
```
ekohl, line 25:
```suggestion
    packages_for_tag = set()
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,70 @@
...
+        if 'koji_tags' in attributes:
+            for koji_tag in attributes['koji_tags']:
+                if tag_name == koji_tag['name']:
+                    tag = koji_tag
+
+            if tag and 'package_base_dir' in attributes:
+                specfile = "%s/%s/%s.spec" % (attributes['package_base_dir'], package, package)
+                scl = tag['scl'] if 'scl' in tag.keys() else None
+                name = get_specfile_name(os.path.join(module.params['directory'], specfile), scl)
+                packages_for_tag.append(name)
```
ekohl, line 39:
```suggestion
                packages_for_tag.add(name)
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,70 @@
...
+
+    for package in packages_for_tag:
+        if package not in packages_in_koji:
+            missing_in_koji.append(package)
+
+    for package in packages_in_koji:
+        if package not in packages_for_tag:
+            missing_in_git.append(package)
+
+    module.exit_json(changed=False, missing_in_git=missing_in_git, missing_in_koji=missing_in_koji)
```
ekohl, line 66:
```suggestion
    module.exit_json(changed=False, missing_in_git=sorted(missing_in_git), missing_in_koji=sorted(missing_in_koji))
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,70 @@
...
+    try:
+        command = ['koji', 'list-pkgs', '--quiet', '--tag', tag_name]
+        packages_in_koji = subprocess.check_output(
+            command,
+            stderr=subprocess.STDOUT,
+            universal_newlines=True
+        )
+
+        packages_in_koji = [item.split(' ')[0] for item in packages_in_koji.split("\n")]
+        packages_in_koji = [package for package in packages_in_koji if len(package) > 0]
```
ekohl, line 50: In Python you don't need `len()` here since empty strings are also `False`. I'd also say that either a line is empty or there's something on it, so you can optimize here. Also, sets are great
```suggestion
        packages_in_koji = {item.split(' ')[0] for item in packages_in_koji.split("\n") if item}
```

This will be different if there are lines where the first char is a space.

obal/data/modules/koji_verify_tag.py
ekohl: I generally don't like changing a variable in this way. First it's a string (after calling `check_output`) and then it becomes a list. That makes it hard to reason about it.
```suggestion
        koji_output = subprocess.check_output(
```

obal/data/modules/koji_find_tags.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+    packages = module.params['packages']
+
+    tags = set()
+
+    for attributes in packages.values():
+        if 'koji_tags' in attributes:
+            for tag in attributes['koji_tags']:
+                tags.add(tag['name'])
```
ekohl, line 24: What's the benefit of looping here instead of bulk adding a set?

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,64 @@
...
+    for (package, attributes) in module.params['packages'].items():
+        tag = None
+
+        if 'koji_tags' in attributes:
+            for koji_tag in attributes['koji_tags']:
+                if module.params['tag'] == koji_tag['name']:
+                    tag = koji_tag
+
+            if tag and 'package_base_dir' in attributes:
+                specfile = "%s/%s/%s.spec" % (attributes['package_base_dir'], package, package)
```
ekohl, line 36:
```suggestion
                specfile = os.path.join(attributes['package_base_dir'], package, "%s.spec".format(package))
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,64 @@
...
+                scl = tag['scl'] if 'scl' in tag.keys() else None
```
ekohl, line 37:
```suggestion
                scl = tag.get('scl')
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,64 @@
...
+                packages_for_tag.add(name)
+
+    try:
+        command = ['list-pkgs', '--quiet', '--tag', module.params['tag']]
+        packages_in_koji = koji(
+            command,
+            executable=module.params['koji_executable']
+        )
+
+        packages_in_koji = set([item.split(' ')[0] for item in packages_in_koji.split("\n") if item])
```
ekohl, line 48: You can use `set(generator)`:

```suggestion
        packages_in_koji = set(item.split(' ')[0] for item in packages_in_koji.split("\n") if item)
```

But shorter (and possibly faster) is set comprehension:

```suggestion
        packages_in_koji = {item.split(' ')[0] for item in packages_in_koji.split("\n") if item}
```

obal/data/modules/koji_find_tags.py
ehelms: I suppose if I put a generator in there to extract the `name` field I could do a bulk add?

ekohl: Ah right, I missed it's extracting the name from `tag` and somehow assumed it was just adding `tag`. In that case, you can use this:
```python
tags += {tag['name'] for tag in attributes['koji_tags']}
```

Though I'm not sure which version states the intent more clearly.

@evgeni any opinion?

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,64 @@
...
+
+            if tag and 'package_base_dir' in attributes:
+                specfile = os.path.join(attributes['package_base_dir'], package, "{}.spec".format(package))
+                scl = tag.get('scl')
+                name = get_specfile_name(os.path.join(module.params['directory'], specfile), scl)
+                packages_for_tag.add(name)
+
+    try:
+        command = ['list-pkgs', '--quiet', '--tag', module.params['tag']]
+        packages_in_koji = koji(
```
ekohl, line 43: I'd still rename this
```suggestion
        koji_output = koji(
```

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,64 @@
...
+        packages_in_koji = {item.split(' ')[0] for item in packages_in_koji.split("\n") if item}
+
+    except subprocess.CalledProcessError as error:
+        module.fail_json(changed=False, msg=error.output)
+
+    missing_in_koji = packages_for_tag - packages_in_koji
+    missing_in_git = packages_in_koji - packages_for_tag
+
+    module.exit_json(
+        changed=False,
```
ekohl, line 57: Should this be true if there is a change?
```suggestion
        changed=missing_in_koji or missing_in_git,
```

Not sure this is semantically entirely correct though since it didn't make a change.

obal/data/roles/verify_tag/tasks/verify.yaml
```diff
@@ -0,0 +1,23 @@
...
+  debug:
+    msg: "{{ missing.missing_in_git }}"
+  when: missing.missing_in_git|length > 0
+
+- name: "Packages missing in koji for {{ tag }}"
+  debug:
+    msg: "{{ missing.missing_in_koji }}"
+  when: missing.missing_in_koji|length > 0
+
+- name: "Print packages missing in koji for {{ tag }}"
```
ekohl, line 20: This doesn't look entirely appropriate, but I'm not sure what it should say.

obal/data/modules/koji_verify_tag.py
ehelms: I can. I just know its a common pattern to capture something and then down below is the same thing but just re-formatting it and using multiple variables to capture that seems overkill here.

obal/data/roles/verify_tag/tasks/verify.yaml
ehelms: In what way?

obal/data/modules/koji_verify_tag.py
ekohl: I think it's a bad pattern. In Python 3 you have type hints which allows static code analysis using tools like [mypy](http://mypy-lang.org/). That's a good practice IMHO and helps you. It also allows IDEs to show more useful suggestions. Sadly, Ansible is incompatible with mypy for now, but it's good to learn to do it the right way:

```console
$ cat test.py 
import subprocess
def test() -> list:
    output = subprocess.check_output('x', universal_newlines=True)
    output = output.split()
    return output
$ mypy test.py 
test.py:6: error: Incompatible types in assignment (expression has type "List[str]", variable has type "str")
test.py:7: error: Incompatible return value type (got "str", expected "List[Any]")
Found 2 errors in 1 file (checked 1 source file)
```

Compare this to:
```console
$ cat test2.py 
import subprocess
def test() -> list:
    output = subprocess.check_output('x', universal_newlines=True)
    result = output.split()
    return result
$ mypy test2.py 
Success: no issues found in 1 source file
```

Now one cool thing it can do:
```console
 cat test3.py 
import subprocess
from typing import List
def test() -> List[str]:
    output = subprocess.check_output('x')
    result = output.split()
    return result
$ mypy test3.py 
test3.py:8: error: Incompatible return value type (got "List[bytes]", expected "List[str]")
Found 1 error in 1 file (checked 1 source file)
```

Note how it caught the missing `universal_newlines` parameter which causes encoding issues.

obal/data/roles/verify_tag/tasks/verify.yaml
ekohl: It doesn't actually print the packages missing, but rather tells you whether it was all ok. `Summarize koji vs sources status` or something feels closer, but it's still not that great.

ekohl:
> @ekohl Does this tag verification concept match what you want ? And is the command formatting OK?

Just wondering if it should be `verify-koji-tag`. It's also too bad we can't use positional arguments for the tags, but I guess that's just one limitation that we need to live with.

ehelms:
That is something I meant to ask @evgeni -- if there was a way in the structure of Obsah to have a command that has positional arguments that are not in the "inventory" ?

Another way around it I considered was having an inventory entry where tags were defined (under a "tags" group) and using that. I was pondering it related to the existence of [koji ansible modules](https://galaxy.ansible.com/ktdreyer/koji_ansible) and defining our tag structure rather than in tool_belt.

evgeni:
> That is something I meant to ask @evgeni -- if there was a way in the structure of Obsah to have a command that has positional arguments that are not in the "inventory" ?

Today? No.

But I think we could rather easily add a flag in the command metadata that says "the params for this playbook (command) aren't hosts (packages)". This would however mean that the playbook would get the params more or less verbatim and would have to process them itself.

ekohl:
The code that generates those is here:
https://github.com/theforeman/obsah/blob/439de5b10a16948d8fd3fbbff0840eaebd93a728/obsah/__init__.py#L300-L305

We can expose the nargs parameter here:
https://github.com/theforeman/obsah/blob/439de5b10a16948d8fd3fbbff0840eaebd93a728/obsah/__init__.py#L307-L309

That allows making tag a positional argument which is stored in an array. I don't really see a way to get the choices without writing Python, but that's probably less important.

ekohl:
> Another way around it I considered was having an inventory entry where tags were defined (under a "tags" group) and using that. I was pondering it related to the existence of [koji ansible modules](https://galaxy.ansible.com/ktdreyer/koji_ansible) and defining our tag structure rather than in tool_belt.

I think that could be nice. Might require some thought on our branching procedure + creating PRs in advance though.

ehelms:
The great irony, this change (https://github.com/theforeman/foreman-packaging/pull/5691) made turning tags into their own group of hosts much much harder as the hosts would need to have a variable in the name and that doesn't work well.

```
tags:
  hosts:
    "foreman-{{ foreman_version }}-rhel7":
      scl: 'tfm'
      dist: '.el7'
      macros:
        foremandist: "{{ foremandist }}"
    "foreman-{{ foreman_version }}-el8":
      dist: '.el8'
      macros:
        foremandist: "{{ foremandist }}"
```

Unless we made the tag entries more generic with configuration below, e.g.

```
tags:
  hosts:
    "foreman-rhel7":
      name: "foreman-{{ foreman_version }}-rhel7"
      scl: 'tfm'
      dist: '.el7'
      macros:
        foremandist: "{{ foremandist }}"
    "foreman-nonscl-rhel7":
      name: "foreman-{{ foreman_version }}-nonscl-rhel7"
      scl: 'tfm'
      dist: '.el7'
      macros:
        foremandist: "{{ foremandist }}"
    "foreman-el8":
      name: "foreman-{{ foreman_version }}-el8"
      dist: '.el8'
      macros:
        foremandist: "{{ foremandist }}"
```

Further downside to hosts entry method, I'm ot sure how to then translate that for easy inclusion as a re-usable unit into the vars for package entries. For example, we define the set of tags associated to a set of packages as a variable. 

```
    koji_tags:
      - name: "foreman-{{ foreman_version }}-rhel7"
        scl: 'tfm'
        dist: '.el7'
        macros:
          foremandist: "{{ foremandist }}"
      - name: "foreman-{{ foreman_version }}-el8"
        dist: '.el8'
        macros:
          foremandist: "{{ foremandist }}"
```

obal/data/modules/koji_find_tags.py
evgeni:
```
TypeError: unsupported operand type(s) for +=: 'set' and 'set'
```

:-)

`tags.update({…})` would work tho.

I'm fine with either form.

obal/data/modules/koji_verify_tag.py
evgeni: I tend to agree with ewoud here

evgeni:
@ehelms so to summarize the discussion above (because I feel a bit lost):

The code in this PR works and the current invocation looks like this:
```console
obal verify-tag --tag foreman-nightly-rhel7
obal verify-tag --all
```

You'd prefer to have this be:
```console
obal verify-tag foreman-nightly-rhel7
obal verify-tag (--)?all
```
but that's not possible due to 1) the way you redefined the version numbering in the manifest 2) us not having a nice way to create a command that takes a positional argument that is not a host?

I can take a stab at 2) but I don't think this should block this PR as the current invocation is "good enough"

evgeni:
Okay, that was easier than expected: https://github.com/theforeman/obsah/pull/6

And no python involved to pull out the data as it gets passed to ansible via `-e blah=blah`

obal/data/playbooks/verify-tag/metadata.obal.yaml
```diff
@@ -0,0 +1,8 @@
+help: Verify packages against a tag
+variables:
+  tag:
+    help: Tag to verify against
```
evgeni, line 4: with https://github.com/theforeman/obsah/pull/6 merged, you can now do:

```suggestion
  tag:
    parameter: tag
    help: Tag to verify against
```

and you'll get a positional argument

mind you, positional arguments are required, so you'll probably want to drop the `verify_all` arg and specialcase on an `all` tag or something

ehelms: Can you release a new Obsah to make testing easier? Also, what does ` specialcase on an all tag or something` mean?

obal/data/playbooks/verify-tag/metadata.obal.yaml
```diff
@@ -0,0 +1,8 @@
...
+  verify_tag_all:
+    action: store_true
+    parameter: --all
+    help: Verify all tags
```
evgeni, line 8:
```suggestion
```

obal/data/roles/verify_tag/tasks/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- include_tasks: "verify.yaml"
+  when: not verify_tag_all
+
+- include_tasks: "all.yaml"
+  when: verify_tag_all
```
evgeni, line 6:
```suggestion
- include_tasks: "verify.yaml"
  when: tag != 'all'

- include_tasks: "all.yaml"
  when: tag == 'all'
```

obal/data/playbooks/verify-tag/metadata.obal.yaml
evgeni: https://github.com/theforeman/obal/pull/243/files#r486301398 + https://github.com/theforeman/obal/pull/243/files#r486301876

ehelms:
Testing locally this seemed to work with the positional args, thanks for tackling that @evgeni ! Now to get these lint failures sorted

evgeni:
I've prepared a new obsah release in https://github.com/theforeman/obsah/pull/7 -- can you please bump the requirement of it in setup.py to >=0.0.2 when it's released?

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,64 @@
...
+                packages_for_tag.add(name)
+
+    try:
+        command = ['list-pkgs', '--quiet', '--tag', module.params['tag']]
+        packages_in_koji = koji(
+            command,
+            executable=module.params['koji_executable']
+        )
+
+        packages_in_koji = {item.split(' ')[0] for item in packages_in_koji.split("\n") if item}
```
ekohl, line 48: Nit: I always like to limit the number of splits if I can. Feels more efficient to me though I never really measured it.
```suggestion
        packages_in_koji = {item.split(' ', 1)[0] for item in packages_in_koji.split("\n") if item}
```

tests/fixtures/help/verify-tag.txt
```diff
@@ -0,0 +1,14 @@
+usage: obal verify-tag [-h] [-v] [-e EXTRA_VARS] [--tag TAG] [--all]
```
evgeni, line 1: this file needs updates now

tests/test_functional.py
```diff
@@ -670,3 +674,22 @@ def test_obal_mock():
...
+
+@obal_cli_test(repotype='upstream')
+def test_obal_verify_tag_all():
+    assert_obal_success(['verify-tag', '--all'])
+
+    expected_log = [
+        'koji list-pkgs --quiet --tag obaltest-nightly-el8',
+        'koji list-pkgs --quiet --tag obaltest-nightly-rhel7'
+    ]
+    assert_mockbin_log(expected_log)
```
evgeni, line 695: these tests need updates now

tests/test_functional.py
```diff
@@ -670,3 +674,22 @@ def test_obal_mock():
         "mock --recurse --chain -r mock/el7.cfg --localrepo {pwd}/mock_builds SRPMs/foo-1.0-1.src.rpm SRPMs/hello-2.10-2.src.rpm"
     ]
     assert_mockbin_log(expected_log)
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_verify_tag():
+    assert_obal_success(['verify-tag', '--tag', 'obaltest-nightly-rhel7'])
```
evgeni, line 681:
```suggestion
    assert_obal_success(['verify-tag', 'obaltest-nightly-rhel7'])
```

tests/test_functional.py
```diff
@@ -670,3 +674,22 @@ def test_obal_mock():
...
+
+    expected_log = ['koji list-pkgs --quiet --tag obaltest-nightly-rhel7']
+    assert_mockbin_log(expected_log)
+
+
+@obal_cli_test(repotype='upstream')
+def test_obal_verify_tag_all():
+    assert_obal_success(['verify-tag', '--all'])
```
evgeni, line 689:
```suggestion
    assert_obal_success(['verify-tag', 'all'])
```

ekohl (changes requested):
This also needs to set the required obsah version to at least 0.0.2

ehelms:
Updated addressing all comments, this needs https://github.com/theforeman/obal/pull/290

ehelms:
Rebased and updated

ehelms:
Current output as of this comment:

```
obal-source verify-koji-tag all

PLAY [localhost] *************************************************************************************************************************************************************************************************************************************************************************
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/all.yaml for localhost

TASK [verify_koji_tag : Find all tags] ***************************************************************************************************************************************************************************************************************************************************
ok: [localhost]
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost

TASK [verify_koji_tag : Verify tag] ******************************************************************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [verify_koji_tag : Packages in Koji but not in git for foreman-client-nightly-el8] **************************************************************************************************************************************************************************************************
ok: [localhost] => 
  msg:
  - jsoncpp
  - qpid-proton

TASK [verify_koji_tag : Verify tag] ******************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [verify_koji_tag : Print packages missing in koji for foreman-client-nightly-el9] ***************************************************************************************************************************************************************************************************
ok: [localhost] => 
  msg: Tag foreman-client-nightly-el9 verified

TASK [verify_koji_tag : Verify tag] ******************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [verify_koji_tag : Print packages missing in koji for foreman-client-nightly-rhel7] *************************************************************************************************************************************************************************************************
ok: [localhost] => 
  msg: Tag foreman-client-nightly-rhel7 verified

TASK [verify_koji_tag : Verify tag] ******************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [verify_koji_tag : Print packages missing in koji for foreman-nightly-el8] **********************************************************************************************************************************************************************************************************
ok: [localhost] => 
  msg: Tag foreman-nightly-el8 verified

TASK [verify_koji_tag : Verify tag] ******************************************************************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [verify_koji_tag : Packages in Koji but not in git for foreman-plugins-nightly-el8] *************************************************************************************************************************************************************************************************
ok: [localhost] => 
  msg:
  - yggdrasil

TASK [verify_koji_tag : Verify tag] ******************************************************************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [verify_koji_tag : Packages in Koji but not in git for katello-nightly-el8] *********************************************************************************************************************************************************************************************************
ok: [localhost] => 
  msg:
  - rubygem-pulp_2to3_migration_client
  - rubygem-redhat_access
  - rubygem-redhat_access_lib

TASK [verify_koji_tag : Packages missing in koji for katello-nightly-el8] ****************************************************************************************************************************************************************************************************************
ok: [localhost] => 
  msg:
  - rubygem-qpid_proton
```

obal/data/modules/koji_find_tags.py
```diff
@@ -0,0 +1,29 @@
+"""
+Find all defined Koji tags in inventory
+"""
+
+from ansible.module_utils.basic import AnsibleModule
+
+def main():
+    """
+    Verify a tag
```
ekohl, line 9: This doesn't match the rest

obal/data/modules/koji_verify_tag.py
```diff
@@ -0,0 +1,64 @@
...
+
+        packages_in_koji = {item.split(' ', 1)[0] for item in koji_output.split("\n") if item}
+
+    except subprocess.CalledProcessError as error:
+        module.fail_json(changed=False, msg=error.output)
+
+    missing_in_koji = packages_for_tag - packages_in_koji
+    missing_in_git = packages_in_koji - packages_for_tag
+
+    module.exit_json(
```
ekohl, line 56: Can this fail `if missing_in_koji or missing_in_git`? Ansible has a way to continue on error, right? Or perhaps the regular failure output is sufficient anyway.

ekohl (commented):
Overall this looks good.

obal/data/modules/koji_verify_tag.py
ehelms: Yes there are two options:

 1) Have the module fail and allow users of the module to ignore errors if they want
 2) Have this module simply report data and then have a task that fails conditionally

I do not know what is more correct, so I am Ok with either strategically.

ekohl: I'd lean to option 1.

ehelms:
New output:

```
obal-source verify-koji-tag all

PLAY [localhost] *************************************************************************************************************************************************************************************************************************************************************************
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/all.yaml for localhost

TASK [verify_koji_tag : Find all tags] ***************************************************************************************************************************************************************************************************************************************************
ok: [localhost]
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost
included: /home/ehelms/workspace/upstream/packaging/obal/obal/data/roles/verify_koji_tag/tasks/verify.yaml for localhost

TASK [verify_koji_tag : Verify tag] ******************************************************************************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => changed=true 
  missing_in_git:
  - jsoncpp
  - qpid-proton
  missing_in_koji: []
  msg: Package differences found

PLAY RECAP *******************************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=8    changed=0    unreachable=0    failed=1    skipped=1    rescued=0    ignored=0   
```

ekohl (commented):
User wise :+1: but CI is unhappy.

ehelms:
@evgeni You did some initial reviews, do you care to have another look? Otherwise I will merge this.

evgeni (approved):

PR #244, opened 2020-04-16
Make KojiCommandError report message and command

ehelms:


obal/data/module_utils/koji_wrapper.py
```diff
@@ -5,6 +5,10 @@
 
 class KojiCommandError(Exception):
     """Raised when Koji command fails"""
+    def __init__(self, message, command):
+        self.message = message
+        self.command = command
+        Exception.__init__(self, message)
```
ekohl, line 11:
```suggestion
        super(KojiCommandError, self).__init__(self, message)
```

ehelms: What's the difference between the two syntax?

ekohl: Your version explicitly calls the constructor from `Exception` while `super()` respects the class inheritance. In your case it doesn't matter that much, but is especially relevant in more complex inheritance. Note that in Python 3 it was simplified to just `super()`.

https://stackoverflow.com/questions/576169/understanding-python-super-with-init-methods has a longer version.

Now that I look at it more awake I think it should even be `super(KojiCommandError, self).__init__(message)`. Python 3 would be `super().__init__(message)`

PR #246, opened 2020-05-01
Drop --best for repoclosure

ehelms:
This option breaks basic repoclosures for EL8 where modules are
involed as repoclosure does not respect module_hotfixes.


ehelms:
Unrelated test failures

PR #260, opened 2020-06-29
migrate to GH actions

evgeni:
based on https://github.com/theforeman/obal/pull/259 this migrates the tests to GH actions which should be faster

draft because:
- [x] missing pypi release on tag


.github/workflows/main.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/checkout@v1
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python }}
+      - name: Install dependencies
+        run: pip install -r requirements-test.txt
+      - name: Install RPM tools
+        run : |
+          curl --create-dirs -o ~/.local/bin/spectool https://pagure.io/rpmdevtools/raw/26a8abc746fba9c0b32eb899b96c92841a37855a/f/spectool.in
```
ekohl, line 27: Would it be easier to use a Fedora container to execute in (since actions allows that)?

.github/workflows/main.yml
```diff
@@ -0,0 +1,50 @@
...
+          curl --create-dirs -o ~/.local/bin/rpmdev-bumpspec https://pagure.io/rpmdevtools/raw/6f387c1deaa5cbed770310e288abde04b17421dc/f/rpmdev-bumpspec
+          curl --create-dirs -o ~/.local/bin/rpmdev-packager https://pagure.io/rpmdevtools/raw/master/f/rpmdev-packager
+          chmod +x ~/.local/bin/spectool
+          chmod +x ~/.local/bin/rpmdev-bumpspec
+          chmod +x ~/.local/bin/rpmdev-packager
+          sudo apt-get install -y git-annex rpm
+      - name: Run tests
+        run: tox
+      - name: Run container tests
+        run: docker build --build-arg VERSION=master -t quay.io/foreman/obal:latest .
```
ekohl, line 37: This doesn't need to be part of the main testing. It builds the same container 3 times.

I'd even argue it's pointless to do this on PR since it only builds master.

.github/workflows/main.yml
evgeni: I'd use CentOS, not Fedora, but yes, I thought about that. Probably better.

.github/workflows/main.yml
evgeni: Yeah, it can be an own step.

Why it builds `master` I can't remember, maybe @ehelms can enlighten us?

ekohl: The PR SHA's are not fetched by default so pypi can't clone them. This is the next best thing. I think it would be better if you could pass in a wheel that was built and install that.

.github/workflows/main.yml
```diff
@@ -0,0 +1,50 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ["2.7", "3.6", "3.7"]
+    steps:
+      - uses: actions/checkout@v1
```
ekohl, line 18: There's also a v2 checkout action. Any reason to stick to v1?

evgeni: no, lazyness, but I added a dependabot config, so it would have fixed that for us soon :)

.github/workflows/main.yml
evgeni: but of course, if I now try to use pip, it fails as the right python version isn't there anymore, whaaaa.

for another day.

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  TERM: 'xterm-256color'
+  PYTEST_ADDOPTS: '--color=yes'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ["2.7", "3.6", "3.7"]
```
ekohl, line 16: I have a slight preference for list style since it's easier to add/remove a single entry since the diff is easier to read.
```suggestion
        python:
          - "2.7"
          - "3.6"
          - "3.7"
```

Perhaps also add 3.8?

.github/workflows/main.yml
evgeni: well, at least that one is fixed now :)

.github/workflows/main.yml
evgeni: I'd like to leave that for another day.

evgeni:
example run: https://github.com/evgeni/obal/runs/869781678?check_suite_focus=true

.github/workflows/main.yml
ekohl: Yes, I think I also ran into the problem that when using a specific container, things like language runtimes (Python, Ruby, probably all others) no longer work. It probably makes sense if they just deploy a tarball that things would break due to linking.

evgeni:
:green_apple: 

still need to do pypi release part

.github/workflows/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          sudo apt-get update && sudo apt-get install -y git-annex rpm
+      - name: Run tests
+        run: tox
+
+  container:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run container tests
+        run: docker build --build-arg REPO=${GITHUB_REPOSITORY} --build-arg VERSION=${GITHUB_REF} -t quay.io/foreman/obal:latest .
```
ekohl, line 44: Nice. I wonder if the tag should be PR, possible with the PR number itself. Right now there's no permission to push, but I think it's good practice to prevent it from even happening.
```suggestion
        run: docker build --build-arg REPO=${GITHUB_REPOSITORY} --build-arg VERSION=${GITHUB_REF} -t quay.io/foreman/obal:pr .
```

Also wondering if you can run the container that was just built and run `obal --help` as a minimal integration test.

evgeni: I couldn't find the PR number in the data we get inside the action. Otherwise I'd prefer to pull `refs/pull/N/merge` as that would have provided the merged result instead of the fork branch (which might be not rebased properly).

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: release
+
+on:
+  push:
+    tags:
+      - obal-[0-9]+.[0-9]+.[0-9]+
```
ekohl, line 6: Since it's a regex, should `.` be escaped?

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+      - name: Install wheel
+        run: pip install wheel
+      - name: Build sdist and wheel
+        run: python setup.py sdist bdist_wheel
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@v1.3.1
```
ekohl, line 24: Is 1.3.1 preferred over a `v1` tag?

evgeni: There is no `v1` in that repo :(

.github/workflows/release.yml
evgeni: it's not a regex. https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

PR #262, opened 2020-07-14
only updat Source lines when source_server is not-empty

evgeni:


obal/data/roles/update_spec_file/tasks/main.yml
```diff
@@ -6,6 +6,7 @@
     replace: '\g<prefix>{{ source_server }}\g<filename>'
   when:
     - source_server is defined
+    - source_server != ""
```
ehelms, line 9: More a question than anything, when does one use `is not` vs `!=` ?

evgeni: I wouldn't know :)

evgeni: In Python, `is` and `==` are different (`is` compares identity, `==` compares values). But I am not sure it matters in Ansible.

evgeni: and ansible-lint told me to write it otherwise anyways :woman_shrugging:

PR #269, opened 2020-09-04
Fix ansible-lint errors

ehelms:


ehelms:
This fails due to newer, Python 3 specific pylint errors:

```
R: 11, 8: Consider using Python 3 style super() without arguments (super-with-arguments)
W: 23, 8: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
```

@evgeni How do I address Python 3 pylint errors without then breaking python 2 pylint?

ekohl:
I think right now you need to disable those lint rules. They're for Python 3-only code bases.

evgeni:
> I think right now you need to disable those lint rules. They're for Python 3-only code bases.

what he said.

ehelms:
I can't figure out how to disable them for Python 3 *only*. If I disable them the way we do elsewhere in the code pylint on Python 2 barfs with an unknown rule.

evgeni (approved):

PR #272, opened 2020-11-03
Fixes #271: Switch diff_package to calculate based on name-version-re…

### LINKED ISSUE #271: diff package fails for packages with epoch
```
'Git version: rubygem-pulpcore_client-1:3.7.1-1.el8'
'Tagged version: rubygem-pulpcore_client-3.7.1-1.el8'
```

ehelms:
…lease

Koji does not report the epoch when looking up package information
so any comparison against a specfile will fail for packages with
an epoch. Given epoch's almost never change, this switches the
diff package check to calculate it based on name-version-release.

### DIFF
--- a/obal/data/roles/diff_package/tasks/koji_tag.yml
+++ b/obal/data/roles/diff_package/tasks/koji_tag.yml
@@ -1,17 +1,17 @@
 ---
 - name: Find local NEVR for package
-  rpm_nevr:
+  rpm_nvr:
     spec_file: "{{ spec_file_path }}"
     scl: "{{ tag.scl | default(omit) }}"
     dist: "{{ tag.dist | default(omit) }}"
     macros: "{{ tag.macros | default(omit) }}"
-  register: package_nevr
+  register: package_nvr
 
 - name: Verify package build exists for tag
   check_koji_build:
     tag: "{{ tag.name }}"
-    nevr: "{{ package_nevr.nevr }}"
-    package: "{{ package_nevr.name }}"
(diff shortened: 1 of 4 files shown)


ehelms:
Updated to fix the tests

ehelms:
Needs https://github.com/theforeman/obal/pull/290

obal/data/roles/diff_package/tasks/koji_tag.yml
```diff
@@ -1,17 +1,17 @@
 ---
 - name: Find local NEVR for package
```
evgeni, line 2: The name is kinda wrong now ;)

obal/data/roles/diff_package/tasks/koji_tag.yml
```diff
@@ -1,17 +1,17 @@
...
     macros: "{{ tag.macros | default(omit) }}"
-  register: package_nevr
+  register: package_nvr
 
 - name: Verify package build exists for tag
   check_koji_build:
     tag: "{{ tag.name }}"
-    nevr: "{{ package_nevr.nevr }}"
-    package: "{{ package_nevr.name }}"
+    nevr: "{{ package_nvr.nvr }}"
```
evgeni, line 13: Should the parameter be also renamed? Or is it used as real nevr still somewhere?

ehelms: Good point, yes.

evgeni (approved):

PR #275, opened 2021-02-11
Re-factor git branch and commit to stand-alone role

ehelms:


ehelms:
@evgeni If you get a moment, this refactor would help some additional work I am doing

obal/data/roles/git_branch_and_commit/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: 'Get updated version from spec'
+  command: "rpmspec --query --queryformat=%{version} --srpm {{ spec_file_path }}"
+  args:
+    chdir: "{{ inventory_dir }}"
+  changed_when: true
```
evgeni, line 6: I'd think this shall be `false` as a query should never *change* something (and it was false in the olde code)

obal/data/roles/git_branch_and_commit/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: 'Get updated version from spec'
+  command: "rpmspec --query --queryformat=%{version} --srpm {{ spec_file_path }}"
```
evgeni, line 3: I wonder if this should be a module, but out of scope here.

obal/data/roles/git_branch_and_commit/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  args:
+    chdir: "{{ inventory_dir }}"
+  changed_when: true
+  register: updated_version
+
+- name: 'Get current git branch'
+  command: "git rev-parse --abbrev-ref HEAD"
+  args:
+    chdir: "{{ inventory_dir }}"
+  changed_when: true
```
evgeni, line 13: false?

obal/data/roles/git_branch_and_commit/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  register: git_branch
+  run_once: true
+
+- name: 'Create git branch'
+  command: "git checkout -b {{ git_branch.stdout }}-update-{{ inventory_hostname }}-{{ updated_version.stdout }}"
+  args:
+    chdir: "{{ inventory_dir }}"
+  run_once: true
+  changed_when: true
```
evgeni, line 22: why the explicit true here?

evgeni (changes requested):

obal/data/roles/git_branch_and_commit/tasks/main.yml
ehelms: I think that would be good as a separate issue.

obal/data/roles/git_branch_and_commit/tasks/main.yml
ehelms: Lint will fail because this is a command without any other indicator of changed. I can add checks for uncommitted files and what branch I am on to make it more idempotent. But I was trying to more re-organize the code rather than so much as change it heavily.

evgeni: interesting, thanks!

evgeni (approved):

PR #276, opened 2021-02-11
Update to ansible-lint 5.0 and add yamllint 

ehelms:


ehelms:
Also includes fixes for container builds

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM centos:7
 
 RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
     yum -y install epel-release git && \
-    yum -y install python2-pip && \
+    yum -y install python-pip && \
```
evgeni, line 5: this *shouldn't* be needed (but also doesn't hurt)

obal/data/playbooks/bump-release/bump-release.yaml
```diff
@@ -12,3 +12,4 @@
       command: "rpmdev-bumpspec {{ spec_file_path | basename }} -c \"{{ changelog }}\""
       args:
         chdir: "{{ spec_file_path | dirname }}"
+      changed_when: false
```
evgeni, line 15: I'd argue this is always changing, as bumpspec will allways bump?

evgeni:
I didn't mean you shall not install `python-pip` at all, just that it doesn't matter if you call it `python-pip` or `python2-pip` ;)

ehelms:
> I didn't mean you shall not install `python-pip` at all, just that it doesn't matter if you call it `python-pip` or `python2-pip` ;)

I see that now!

evgeni (approved):

PR #278, opened 2021-03-05
Only copy Source and Patch files when building an SRPM

ehelms:
Otherwise, rpmlint errors can interfere:

```
TASK [build_srpm : Build SRPM] ***********************************************************************************************************************************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: FileNotFoundError: [Errno 2] No such file or directory: '/home/ehelms/workspace/upstream/foreman-packaging/packages/katello//qpid-proton/line'
fatal: [qpid-proton]: FAILED! => changed=false 
  module_stderr: |-
    Traceback (most recent call last):
      File "/home/ehelms/.ansible/tmp/ansible-tmp-1614952672.0688179-3192245-280641090665047/AnsiballZ_srpm.py", line 102, in <module>
        _ansiballz_main()
      File "/home/ehelms/.ansible/tmp/ansible-tmp-1614952672.0688179-3192245-280641090665047/AnsiballZ_srpm.py", line 94, in _ansiballz_main
        invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
      File "/home/ehelms/.ansible/tmp/ansible-tmp-1614952672.0688179-3192245-280641090665047/AnsiballZ_srpm.py", line 40, in invoke_module
        runpy.run_module(mod_name='ansible.modules.srpm', init_globals=None, run_name='__main__', alter_sys=True)
      File "/usr/lib64/python3.9/runpy.py", line 210, in run_module
        return _run_module_code(code, init_globals, run_name, mod_spec)
      File "/usr/lib64/python3.9/runpy.py", line 97, in _run_module_code
        _run_code(code, mod_globals, init_globals,
      File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
        exec(code, run_globals)
      File "/tmp/ansible_srpm_payload_tbtqisp4/ansible_srpm_payload.zip/ansible/modules/srpm.py", line 137, in <module>
      File "/tmp/ansible_srpm_payload_tbtqisp4/ansible_srpm_payload.zip/ansible/modules/srpm.py", line 100, in main
      File "/tmp/ansible_srpm_payload_tbtqisp4/ansible_srpm_payload.zip/ansible/modules/srpm.py", line 53, in copy_sources
      File "/usr/lib64/python3.9/shutil.py", line 418, in copy
        copyfile(src, dst, follow_symlinks=follow_symlinks)
      File "/usr/lib64/python3.9/shutil.py", line 264, in copyfile
[...]

PR #279, opened 2021-03-10
Add srpm source location

ehelms:
This fills a gap in the srpm.py module that was added a few months back. A source location and source type can now be specified that will download sources into the buildroot. This mimcs the behavior that tito used to provide in this area and is focused on supporting our primary use of downloading nightly sources from Jenkins. This will require changes to the package manifest to support it:

```
    foreman:
      releasers:
        - "{{ nightly_releaser }}"
      build_package_tito_releaser_args: "{{ nightly_package_tito_releaser_args }}"
      nightly_package_tito_releaser_args:
        - "jenkins_job=foreman-develop-source-release"
        - "jenkins_url=https://ci.theforeman.org"
      build_package_tito_builder: "tito.builder.FetchBuilder"
      source_location: https://ci.theforeman.org/job/foreman-develop-source-release
      source_system: jenkins
```

Note the last two new parameters. I am not sold on the naming convention of the variables, so feel free to offer new ideas. The tito arguments will have to remain for a period of time as those are still used by the koji build functionality (https://github.com/theforeman/obal/pull/230 aims to close that gap).

This is built on top of https://github.com/theforeman/obal/pull/278 which touches the same file.

### DIFF
--- a/obal/data/modules/srpm.py
+++ b/obal/data/modules/srpm.py
@@ -5,11 +5,16 @@ Build SRPM
 import shutil
 import os
 import subprocess
+from zipfile import ZipFile
 from contextlib import contextmanager
-from tempfile import mkdtemp
+from tempfile import mkdtemp, TemporaryFile
+
+from ansible.module_utils.six.moves.urllib.request import urlopen # pylint:disable=import-error,no-name-in-module
+from ansible.module_utils.six.moves.urllib.error import HTTPError # pylint:disable=import-error,no-name-in-module
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.obal import run_command, get_specfile_sources # pylint:disable=import-error,no-name-in-module
+
 
 @contextmanager
 def chdir(directory):
@@ -54,6 +59,32 @@ def copy_sources(spec_file, package_dir, sources_dir):
             run_command(["git-annex", "lock", "--force"])
 
 
+def fetch_remote_sources(source_location, source_system, sources_dir):
+    """
+    Copy RPM sources from a remote source like Jenkins to rpmbuild environment
+    """
+    source_system_urls = {
+        'jenkins': '{}/lastSuccessfulBuild/artifact/*zip*/archive.zip',
+    }
+
+    url = source_system_urls[source_system].format(source_location)
+    request = urlopen(url)
+
+    try:
+        archive = TemporaryFile()
+        archive.write(request.read())
+
+        with ZipFile(archive) as zip_file:
+            for zip_info in zip_file.infolist():
+                if zip_info.filename[-1] == '/':
+                    continue
+
+                zip_info.filename = os.path.basename(zip_info.filename)
+                zip_file.extract(zip_info, sources_dir)
+    finally:
+        archive.close()
+
+
 def main():
     """
     Build a package using tito
@@ -63,12 +94,16 @@ def main():
             package=dict(type='str', required=False),
             scl=dict(type='str', required=False),
             output=dict(type='path', required=False),
+            source_location=dict(type='str', required=False),
+            source_system=dict(type='str', required=False),
         )
     )
 
     package = module.params['package']
     output = module.params['output']
     scl = module.params['scl']
+    source_location = module.params['source_location']
+    source_system = module.params['source_system']
 
     spec_file = os.path.join(package, '%s.spec' % os.path.basename(package))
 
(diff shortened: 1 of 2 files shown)


evgeni:
@ehelms would you mind rebasing and fixing the "  C: 59, 0: Missing function or method docstring (missing-function-docstring)" pylint error?

obal/data/modules/srpm.py
```diff
@@ -54,6 +56,27 @@ def copy_sources(spec_file, package_dir, sources_dir):
             run_command(["git-annex", "lock", "--force"])
 
 
+def fetch_remote_sources(source_location, source_system, sources_dir):
+    """
+    Copy RPM sources from a remote source like Jenkins to rpmbuild environment
+    """
+    if source_location and source_system:
+        url = "%s/lastSuccessfulBuild/artifact/*zip*/archive.zip" % (source_location)
```
evgeni, line 64: This structure implies `source_system == 'jenkins'`, but you never check that.

A possible solution could be having a global dict with systems and their url structure and use it here:

```python
SOURCE_SYSTEM_URLS = {
    'jenkins': '{}/lastSuccessfulBuild/artifact/*zip*/archive.zip',
}
…
def fetch_remote_sources(source_location, source_system, sources_dir):
    …
    url = SOURCE_SYSTEM_URLS[source_system].format(source_location)
```

And then you'd catch a `KeyError` below and tell the user "unknown source system"

obal/data/modules/srpm.py
```diff
@@ -54,6 +56,27 @@ def copy_sources(spec_file, package_dir, sources_dir):
             run_command(["git-annex", "lock", "--force"])
 
 
+def fetch_remote_sources(source_location, source_system, sources_dir):
+    """
+    Copy RPM sources from a remote source like Jenkins to rpmbuild environment
+    """
+    if source_location and source_system:
```
evgeni, line 63: technically you'd need to check both for `is not None` only, and that's a weird API (as the params are required).

I'd personally skip that check here, and only call the method from `main()` if the vars are set *there*

evgeni (changes requested):
comments on the python code, but happy with the overall "this is how it shall work in obal"

variable names are fine too

obal/data/modules/srpm.py
ehelms: Excellent. I've implemented this tactic.

ehelms:
To get python 2.7 I had to change my approach and dance around with urllib being different in Py2 and Py3.

obal/data/modules/srpm.py
```diff
@@ -5,8 +5,15 @@
...
+from zipfile import ZipFile
 from contextlib import contextmanager
-from tempfile import mkdtemp
+from tempfile import mkdtemp, TemporaryFile
+
+try:
+    from urllib.request import urlopen
+    from urllib.error import HTTPError
+except ImportError:
+    from urllib2 import urlopen, HTTPError
```
evgeni, line 16: Ansible vendors [`six`](https://six.readthedocs.io/), and that has [`six.moves`](https://six.readthedocs.io/#module-six.moves) to take care of these games.

So you could do smth like:

```python
from ansible.module_utils.six.moves.urllib.request import urlopen
from ansible.module_utils.six.moves.urllib.error import HTTPError
```

(untested)

But you don't have to, just a comment ;)

obal/data/modules/srpm.py
```diff
@@ -54,6 +61,32 @@ def copy_sources(spec_file, package_dir, sources_dir):
             run_command(["git-annex", "lock", "--force"])
 
 
+def fetch_remote_sources(source_location, source_system, sources_dir):
+    """
+    Copy RPM sources from a remote source like Jenkins to rpmbuild environment
+    """
+    source_system_urls = {
```
evgeni, line 68: I thought more of a global, static var (`SOURCE_SYSTEM_URLS`), but wouldn't mind either way here.

obal/data/modules/srpm.py
```diff
@@ -54,6 +61,32 @@ def copy_sources(spec_file, package_dir, sources_dir):
...
+        'jenkins': '{}/lastSuccessfulBuild/artifact/*zip*/archive.zip',
+    }
+
+    url = source_system_urls[source_system].format(source_location)
+    request = urlopen(url)
```
evgeni, line 73: Ansible has `open_url`, `fetch_url` and `fetch_file` in https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py which could be used instead, but I don't think we buy anything by using them here.

evgeni (approved):
left a few comments what could be done *differently* without really thinking it'd be *better* ;-)

obal/data/modules/srpm.py
ehelms: I was worried about scope creep, and I figured if I went this route and there was a reason to go with a global static variable you'd point it out to me :)

PR #283, opened 2021-03-25
Return the changelog from the changelog module

ehelms:
Additionally changes the changelog module parameter from changelog
to entry to better reflect what the parameter being supplied is
and reduce the occurrances of the word changelog. The intent is for
changelog to represent the whole changelog and entry to represent
an item in the changelog.

### DIFF
--- a/obal/data/modules/changelog.py
+++ b/obal/data/modules/changelog.py
@@ -35,12 +35,12 @@ def main():
     module = AnsibleModule(
         argument_spec=dict(
             spec=dict(required=True),
-            changelog=dict(default='- rebuilt')
+            entry=dict(required=False)
         )
     )
 
     spec = module.params['spec']
-    changelog = module.params['changelog']
+    entry = module.params['entry']
 
     user = subprocess.check_output(['rpmdev-packager'], universal_newlines=True).strip()
     evr = get_specfile_evr(spec)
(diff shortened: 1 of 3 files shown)


evgeni:
Do I read it right, one can run the module w/o `entry` now, and just gets the old changelog back?

obal/data/modules/changelog.py
```diff
@@ -50,22 +50,27 @@ def main():
 
     changed = False
 
-    if not changelog.startswith('-'):
-        changelog = '- ' + changelog
+    if entry is not None and not entry.startswith('-'):
```
ekohl, line 53: It's good enough to check if it's trueish. This also catches empty strings but that's ok because an empty string never starts with `-`.
```suggestion
    if entry and not entry.startswith('-'):
```

ehelms:
> Do I read it right, one can run the module w/o `entry` now, and just gets the old changelog back?

That is one ~~consequence~~ feature of this change. Would you want the changelog command to print the changelog if no entry is supplied?

evgeni:
Nah, just curious what this change does :)

ehelms:
The first use of the return will be making updates to the specfile via something like `gem2rpm` and writing back the original changelog.

evgeni (approved):

PR #285, opened 2021-03-29
include the prerelease in the generated changelog

evgeni:


ekohl (commented):
Thanks! I'm not exactly sure but can we also correct the commit message of `obal update`? I think that's also generated. I think that's this line:
https://github.com/theforeman/obal/blob/6920b6b0a0bbf15c67080896e4af050d91d85827/obal/data/roles/git_branch_and_commit/tasks/main.yml#L3

evgeni:
Yes, but I'd prefer to defer that to a later PR as I'd love to make more changes to that portion of the code.

ekohl (approved):

ekohl:
(tests are still running)

PR #290, opened 2021-06-16
Fix consider-using-with pylint error in srpm module

ehelms:


ekohl (changes requested):
This isn't really fixing it. Also, why does it need to be done globally? I'd prefer a control statement inside the file to silence it locally if we can't fix it.

I wasn't really sure where it was, but this looks like the place: https://github.com/theforeman/obal/pull/291.

ehelms:
The reason this had to be done globally is Python 2 vs Python 3. This check is only present for Pylint versions that support Python 3 and it throws errors if you attempt to in-line disable a rule that does not exist.

PR #293, opened 2021-06-17
Change package_whitelist_check to package_tag_check

ehelms:
We originally named this whitelist because that is the term tito
used to declaring the set of packages for a tag. This steps towards
using Koji directly by talking of checking if a package exists for
a tag.


ehelms:
This is a change split off from https://github.com/theforeman/obal/pull/230 as its stand-alone. This will need one CI update: https://github.com/theforeman/jenkins-jobs/pull/72

PR #178, opened 2019-03-14
Add a build_source module

ekohl:
Currently we have a tito extension in foreman-packaging. The idea is expand this module to replace that and support building just by pointing at a git checkout. This is easy for development but also makes it easy to write a Jenkins job to build a scratch build of a PR.

PR #295, opened 2021-08-31
Implement local source fetching

ekohl:
This adds the ability to fetch sources by using:
(bundle exec) rake pkg:generate_source

This is implemented in the Foreman repositories and makes it easier to run local scratch builds.

It also fixes handling of remote sources and a TypeError.

The tests currently fail, but at least I've pushed my changes so I don't lose them.

### DIFF
--- a/obal/data/modules/srpm.py
+++ b/obal/data/modules/srpm.py
@@ -64,6 +64,22 @@ def copy_sources(spec_file, package_dir, sources_dir):
             run_command(["git-annex", "lock", "--force"])
 
 
+def fetch_local_sources(source_location, source_system, sources_dir):
+    if source_system == 'local-bundle-rake':
+        command = ['bundle', 'exec', 'rake', 'pkg:generate_source']
+    elif source_system == 'local-rake':
+        command = ['rake', 'pkg:generate_source']
+    else:
+        raise ValueError("Unknown source system %s" % source_system)
+
+    with chdir(source_location):
+        output = run_command(command)
+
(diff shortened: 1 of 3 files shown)


ehelms:
This initially feels like over reach to me. As in, something should have the ability to build or generate local source files but that should be a separate module or role that can then hand that off to the `srpm` module as inputs. That would align it with the remote sources that is simply fetching sources.

ekohl:
The thing I want to add too is to retrieve the commit and include that in the `prerelease` macro. Today we have that for regular nightly packages (see https://yum.theforeman.org/nightly/el8/x86_64/), but it's not there if you'd do a local mock build. I still consider that we don't have this easily available anymore in foreman-packaging a major regression from where we were.

These instructions to build from a local directory (and for srpm) do not work anymore: https://github.com/theforeman/foreman-packaging/tree/rpm/develop#with-mock.

ehelms:
I think you were on the right track with https://github.com/theforeman/obal/pull/178 as a dedicated module to handle building, that returns a path to the built source. That built source can then be fed into the `srpm` module as the `source_location`. So a parameter could trigger to build a local source based on a path (e.g. `--build-source-dir`) and that could trigger a block that builds the source and then overrides `source_location` to point at the local location for the source and `source_system` would be `local`.

This could even open the door for allowing a source_system that is github, source_location being the git repository, and source_ref that could handle building a PR. The build module you wrote could then fetch and build in a temporary directory and feed that to srpm module.

PR #302, opened 2022-05-27
Fix pylint errors, disable f-string

ehelms:
I am opting to intentionally disable f-string check for now as it is triggered in places where more care must be taken in formatting of strings due to how they are output to the command line.


obal/data/modules/changelog.py
```diff
@@ -45,7 +45,7 @@ def main():
     user = subprocess.check_output(['rpmdev-packager'], universal_newlines=True).strip()
     evr = get_specfile_evr(spec)
 
-    with open(spec) as spec_file:
+    with open(spec, encoding="utf-8") as spec_file:
```
evgeni, line 48: open (encoding=...) is Python 3 only, I wonder why this didn't trip the CI?!

evgeni (changes requested):

ehelms:
We still using Python 2!?

obal/data/modules/changelog.py
ehelms: Going to disable that for now.

evgeni:
> We still using Python 2!?

/me looks at all the EL7 based Jenkins nodes

evgeni (approved):

PR #303, opened 2022-05-27
Add koji build support

ehelms:
Requires -- ~~https://github.com/theforeman/obal/pull/302~~

This is an evolution of https://github.com/theforeman/obal/pull/230 where this PR adds the functionality and the ability to enable this functionality directly through a flag. All existing projects using this will continue to use tito as before. But projects can opt in to testing/using builds directly through koji by setting:

```
build_package_use_koji_build: true
```

### DIFF
--- a/obal/data/modules/koji_build.py
+++ b/obal/data/modules/koji_build.py
@@ -1,0 +1,69 @@
+"""
+Release a package to koji
+"""
+
+import re
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.koji_wrapper import koji, KojiCommandError # pylint:disable=import-error,no-name-in-module
+from ansible.module_utils.obal import get_whitelist_status  # pylint:disable=import-error,no-name-in-module
+
+
+def main():
+    """
+    Build a package in koji
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            srpm=dict(type='path', required=True),
+            scratch=dict(type='bool', required=False, default=False),
+            tag=dict(type='str', required=True),
+            nevr=dict(type='str', required=True),
+            package=dict(type='str', required=True),
+            tag_check=dict(type='bool', required=False, default=True),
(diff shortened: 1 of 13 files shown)


obal/data/modules/koji_build.py
```diff
@@ -0,0 +1,69 @@
...
+
+    tag = module.params['tag']
+    scratch = module.params['scratch']
+    srpm = module.params['srpm']
+    nevr = module.params['nevr']
+    package = module.params['package']
+    koji_executable = module.params['koji_executable']
+
+    if module.params['tag_check']:
+        if not get_whitelist_status(module.params['koji_executable'], tag, package):
```
evgeni, line 36: `module.params['koji_executable']` can be `None`, and I don't see any assertions for `koji list-pkgs` in the log below?

evgeni: (also, `get_whitelist_status` doesn't use the `koji` wrapper we have in `obal/data/module_utils/koji_wrapper.py`, which would have saved you from this issue)

obal/data/modules/koji_build.py
```diff
@@ -0,0 +1,69 @@
...
+    if not scratch:
+        try:
+            command = ['latest-build', '--quiet', tag, package]
+            koji_output = koji(command, koji_executable)
+            build = koji_output.split(' ')[0]
+
+            if build == nevr:
+                module.exit_json(changed=False, build=build)
+        except KojiCommandError as error:
+            module.fail_json(changed=False, msg=error, command=command)
```
evgeni, line 48: this section is not tested :)

ehelms:
This now includes updated tests

obal/data/modules/koji_build.py
evgeni, line 36:
```suggestion
        if not get_whitelist_status(koji_executable, tag, package):
```

you have that var, so let's use it :)

evgeni (approved):

PR #304, opened 2022-05-31
Use splitlines instead of split for getting evr from changelog

ehelms:
This is breaking CI tasks presently:

https://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-rpm-pr-test/detail/foreman-packaging-rpm-pr-test/7834/pipeline/

### DIFF
--- a/obal/data/module_utils/obal.py
+++ b/obal/data/module_utils/obal.py
@@ -40,7 +40,7 @@ def get_changelog_evr(specfile):
         specfile
     ]
     evr = subprocess.check_output(cmd, universal_newlines=True)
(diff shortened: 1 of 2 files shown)


obal/data/module_utils/obal.py
```diff
@@ -40,7 +40,7 @@ def get_changelog_evr(specfile):
         specfile
     ]
     evr = subprocess.check_output(cmd, universal_newlines=True)
-    return evr.split('\n', maxsplit=1)[0].split(" ")[-1]
+    return evr.split("\n")[0].split(" ")[-1]
```
evgeni, line 43: How about `evr.splitlines()` instead?

PR #305, opened 2022-06-01
Add ability to check if package exists in Koji

ehelms:
Previously, the check_koji_build only allowed checking if a package
build existed within a tag, this allows checking if the build exists
in Koji anywhere.

### DIFF
--- a/obal/data/modules/check_koji_build.py
+++ b/obal/data/modules/check_koji_build.py
@@ -11,7 +11,7 @@ def main():
     """
     module = AnsibleModule(
         argument_spec=dict(
-            tag=dict(type='str', required=True),
+            tag=dict(type='str', required=False),
             nevr=dict(type='str', required=True),
             package=dict(type='str', required=True),
             koji_executable=dict(type='str', required=False)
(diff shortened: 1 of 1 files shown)


ehelms:
This is a small step towards adding the ability to detect if a build exists and then tag it in to a tag.

obal/data/modules/check_koji_build.py
```diff
@@ -23,13 +23,24 @@ def main():
...
         build = koji(command, koji_executable)
-        build = build.split(' ')[0]
 
-        module.exit_json(changed=False, tagged_version=build, exists=(nevr == build))
+        if tag:
+            build = build.split(' ')[0]
+
+            module.exit_json(changed=False, tagged_version=build, exists=(nevr == build))
+        else:
+            output = koji(command, koji_executable)
```
evgeni, line 39: this executes `koji(command)` a second time (the first one in line 32).

I'd argue for better readability to fold the `if/else` from lines 26-29 with the one here too

evgeni:
I'd love to see tests added for this!

Love the feature!

ehelms:
> I'd love to see tests added for this!
> 
> Love the feature!

Hrmm, so today we do not have direct module tests, and this new functionality is not actually used until https://github.com/theforeman/obal/pull/306/files#diff-3fa87caeaa20d79dd42a779614b1c8a7b57acd53cbe929aa3cc55274f6223eb2R10 . So adding new tests is hard... I can close this an incorporate this change only in #306 ?

evgeni (approved):
let's test it in #306!

PR #306, opened 2022-06-01
Tag package build into tag if it already exists

ehelms:


evgeni:
guess this needs a rebase now :)

obal/data/modules/check_koji_build.py
```diff
@@ -23,23 +23,35 @@ def main():
...
 
+    command = ['buildinfo', nevr]
     try:
-        if tag:
-            command = ['latest-build', '--quiet', tag, package]
-            build = koji(command, koji_executable)
-            build = build.split(' ')[0]
+        output = koji(command, koji_executable)
+    except KojiCommandError as error:
+        module.fail_json(changed=False, msg=error.message, command=error.command)
```
evgeni, line 30: when the build doesn't exist, `koji buildinfo` exits non-zero, which will trigger this Exception, even if you don't want it:

```console
% koji buildinfo hello-123-1.el7; echo $?       
No such build: hello-123-1.el7

1
```

obal/data/modules/check_koji_build.py
```diff
@@ -23,23 +23,35 @@ def main():
...
 
-            module.exit_json(changed=False, exists=exists)
+    if tag:
+        exists_for_tag = False
 
-    except KojiCommandError as error:
-        module.fail_json(changed=False, msg=error.message, command=error.command)
+        for line in output.split("\n"):
+            if line.startswith("Tags:"):
+                exists_for_tag = tag in line
```
evgeni, line 39: while probably uncommon, this would match if we configure `tag` as `tag1` but the build is tagged into `tag1-candidate`

IMHO more correct:
```python
if line.startswith("Tags: "):
    tags = line.split()[1:]
    exists_for_tag = tag in tags
```

obal/data/modules/check_koji_build.py
```diff
@@ -23,23 +23,35 @@ def main():
...
 
+            try:
+                build = koji(command, koji_executable)
+            except KojiCommandError as error:
+                module.fail_json(changed=False, msg=error.message, command=error.command)
+
+            build = build.split(' ')[0]
+            module.exit_json(changed=False, exists=exists, tagged_version=build, exists_for_tag=False)
+        else:
+            module.exit_json(changed=False, exists=exists, tagged_version=build, exists_for_tag=True)
```
evgeni, line 52: there is no `build` in the `else` branch defined.
you could use `nevr` here, as you checked that it's the tagged version?

ehelms:
I had forgot will also need https://github.com/theforeman/obal/pull/272 to pull into this code

evgeni:
Request for Rebase!

obal/data/roles/build_package/tasks/koji_build.yml
```diff
@@ -1,50 +1,76 @@
...
-    koji_tasks: "{{ koji_build_tasks.tasks }}"
-
-- name: 'Download task results'
-  include_tasks: download_rpms.yml
-  when: build_package_download_rpms|bool
+  register: build_exists
+  when: not build_package_scratch
+
+- debug:
+    msg: "{{ build_exists }}"
```
evgeni, line 20: should this be a nicer message (or removed?)

evgeni (approved):

obal/data/roles/build_package/tasks/koji_build.yml
```diff
@@ -1,50 +1,76 @@
...
+    - build_exists is not skipped
+    - not build_exists.exists_for_tag
+    - build_exists.exists
+    - not build_package_scratch
+  block:
+    - name: Tag package build into Koji tag
+      koji_tag_package:
+        tag: "{{ tag.name }}"
+        nvr: "{{ package_nvr.nvr }}"
+        koji_executable: "{{ koji_executable|default('koji') }}"
```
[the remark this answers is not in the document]
ehelms: Ja, will drop

obal/data/roles/build_package/tasks/koji_build.yml
evgeni: this is the debug output of `diff_package` from https://github.com/theforeman/obal/blob/6f66c0caac0409488312a8062ef77fc4e09138d3/obal/data/roles/diff_package/tasks/koji_old.yml#L31-L36

(well, or your edited version of that)

but the comment here is on he `build_package` role and the variable is coming from the `register` step before.

ehelms: Oh duh, I forgot to pass the flag when generating the output:

```
TASK [build_package : debug] *************************************************************************************************************************************************************************************************************************************************************
ok: [rubygem-deep_cloneable] => 
  msg:
    changed: false
    exists: true
    exists_for_tag: true
    failed: false
    tagged_version: rubygem-deep_cloneable-3.0.0-4.el8
```

evgeni: nah, works for me!

PR #307, opened 2022-06-01
Swap container to be stream8 based

ehelms:


Dockerfile
```diff
@@ -1,20 +1,23 @@
-FROM centos:7
+FROM centos:stream8
```
evgeni, line 1:
```suggestion
FROM quay.io/centos/centos:stream8
```

Dockerfile
```diff
@@ -1,20 +1,23 @@
...
 
 RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
-    yum -y install epel-release git && \
-    yum -y install python2-pip && \
+    yum -y install git glibc-langpack-en epel-release && \
```
evgeni, line 4: we shouldn't need `epel-release` anymore (pip is in centos these days)

ehelms: Truth. Koji is in epel though.

evgeni: Huh, yeah.

Dockerfile
```diff
@@ -1,20 +1,22 @@
-FROM centos:7
+FROM quay.io/centos/centos:stream8
 
 RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
-    yum -y install epel-release git && \
-    yum -y install python2-pip && \
+    yum -y install git glibc-langpack-en epel-release && \
+    yum -y install python3-pip && \
```
evgeni, line 5: You could fold this into the previous line, saving one yum transaction. But not tooo important.

evgeni (approved):

PR #311, opened 2022-06-22
Cleaning the Dockerfile to remove some layers

Odilhao:


evgeni:
This seems to have unnecessary changes (actually reverting things we committed today)

Odilhao:
I had to do this changes to speed the container creation locally, will this revert the git-annex change? :thinking:

evgeni:
Interesting, now the diff renders correctly.

Odilhao:
Rebased against the latest commit on master, I don't know why the diff showed like that

ehelms (approved):

evgeni:
(the fact that this is necessary underlines my disliking of the containerfile syntax, but oh well, here we are.

Issue #318, opened 2022-08-19
non-tito builds don't contain %{dist} and %{foremandis} macros in SRPM name


evgeni:
https://github.com/theforeman/foreman-packaging/pull/8364 (3.3 tito-based build) links in checks to "koji/rubygem-foreman_webhooks-3.0.4-1.fm3_3.el8.src.rpm"
https://github.com/theforeman/foreman-packaging/pull/8362 (nightly, non-tito) links to "koji/rubygem-foreman_webhooks-3.0.4-1.src.rpm"

This is confusing :)

ehelms:
The design idea with build SRPM is that it only needs to build the SRPM once and then send that up to Koji or mock. There are two approaches we can take:

 1) Change it so that it generates an SRPM per dist and foremandist
 2) Change our task info logic to look at child tasks and pluck out hte value form the rebuildSRPM task

evgeni:
I think 2 is the cleaner solution given the (now explained and found good) design :)

evgeni:
@ehelms another interesting side-effect: nightly builds don't have their nightly tag when built from a PR: http://koji.katello.org/koji/buildinfo?buildID=72915

(cc @ekohl )

ehelms:
@evgeni This was created back when we used Koji and was primarily a CI issue where the individual jobs per dist were not clear, now with Copr we display it as:

```
@theforeman-bot
copr/foreman_ygg_worker-opensuse-leap-15.5-x86_64 — Copr build
[Details](https://copr.fedorainfracloud.org/coprs/build/8162168)
@theforeman-bot
copr/foreman_ygg_worker-rhel-7-x86_64 — Copr build
[Details](https://copr.fedorainfracloud.org/coprs/build/8162167)
@theforeman-bot
copr/foreman_ygg_worker-rhel-8-x86_64 — Copr build
[Details](https://copr.fedorainfracloud.org/coprs/build/8162166)
@theforeman-bot
copr/foreman_ygg_worker-rhel-9-x86_64 — Copr build
```

Where the chroot is used for each one from Copr. This would not include the foremandist aspect however. The builds as listed (https://copr.fedorainfracloud.org/coprs/g/theforeman/plugins-nightly-staging/builds/) lack the dist and foremandist because that represents the SRPM that was uploaded. If you think we should still tackle this and ensure it includes dist/foremandist in the SRPM then I'll tackle making this change:

> Change it so that it generates an SRPM per dist and foremandist

[...]

PR #320, opened 2022-08-30
Allow passing macros to SRPM build

ehelms:
Refs #318 

This aligns better with any provided macros by the project and allows for passing a nightly macro.

### DIFF
--- a/obal/data/modules/srpm.py
+++ b/obal/data/modules/srpm.py
@@ -90,6 +90,7 @@ def main():
         argument_spec=dict(
             package=dict(type='str', required=False),
             scl=dict(type='str', required=False),
+            macros=dict(type='dict', required=False, default={}),
             output=dict(type='path', required=False),
             source_location=dict(type='str', required=False),
             source_system=dict(type='str', required=False),
@@ -99,6 +100,7 @@ def main():
     package = module.params['package']
     output = module.params['output']
     scl = module.params['scl']
+    macros = module.params['macros']
     source_location = module.params['source_location']
     source_system = module.params['source_system']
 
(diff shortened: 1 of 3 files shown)


obal/data/modules/srpm.py
```diff
@@ -134,6 +136,10 @@ def main():
         if scl:
             command += ['--define', 'scl %s' %  scl]
 
+        if macros is not None:
```
ekohl, line 139: If the default is `{}`, is this ever None? I kind of forgot what `argument_spec` really does.

ekohl (commented):
Makes sense.

obal/data/modules/srpm.py
evgeni: If the `default` is `{}` then this is never `None`, no. It's only `None` for fields that don't have a default and are `required=False`.

obal/data/modules/srpm.py
```diff
@@ -83,7 +83,7 @@ def fetch_remote_sources(source_location, source_system, sources_dir):
                 zip_file.extract(zip_info, sources_dir)
 
 
-def build_srpm(module, package, base_dir, sources_dir, scl=None):
+def build_srpm(module, package, base_dir, sources_dir, scl=None, macros=None):
```
evgeni, line 86: pylint hates this code:
```
  R: 86, 0: Too many arguments (6/5) (too-many-arguments)
  R: 86, 0: Too many positional arguments (6/5) (too-many-positional-arguments)
```

That said, I don't like that check and would probably just disable it globally.

evgeni (approved):

ehelms:
I am running into pylint not being the same across the versions of Python and these two rules not existing on 3.6/3.7/3.8. I don't recall having ever seen a clean way to handle this situation.

evgeni:
> I am running into pylint not being the same across the versions of Python and these two rules not existing on 3.6/3.7/3.8. I don't recall having ever seen a clean way to handle this situation.

That's "fine". We had pylint failing on 3.6-3.8 python before this change already. As long as 3.9+ is happy, I'd just merge this.

PR #324, opened 2022-10-11
Adding the git_branching option

Odilhao:
This will be used for pulpcore branching, the --commit make it hard to push/sync during branching, when we have at least 25+ package updates.

For this I'm adding a `--branching` to keep all commits in the same branch

### DIFF
--- a/obal/data/playbooks/update/metadata.obal.yaml
+++ b/obal/data/playbooks/update/metadata.obal.yaml
@@ -8,9 +8,13 @@ help: |
 
     obal update mypackage --version 3.4.5
 
-  To update a package to a new version and create a git branch for the change from the command line:
+  To update a package to a new version and create a commit for the change from the command line:
 
     obal update mypackage --version 3.4.5 --commit
+
+  To update a package to a new version and create a git branch for the change from the command line:
+
+    obal update mypackage --version 3.4.5 --branch
 
   Updating a package that is taken from an upstream repository assumes that the package_manifest already defines the upstream location. Thus, the command line to update to a newer version that is available upstream automatically is:
 
@@ -25,5 +29,8 @@ variables:
   changelog:
     help: Optionally set the changelog. When unspecified, an entry will be generated
   commit:
+    help: When true, commits the update changes to it into the current branch
+    action: store_true
+  branch:
     help: When true, creates a git branch and commits the update changes to it.
     action: store_true
--- a/tests/fixtures/help/update.txt
+++ b/tests/fixtures/help/update.txt
@@ -1,5 +1,6 @@
 usage: obal update [-h] [-v] [-e EXTRA_VARS] [--changelog CHANGELOG]
-                   [--commit] [--prerelease PRERELEASE] [--release RELEASE]
+                   [--branch] [--commit]
+                   [--prerelease PRERELEASE] [--release RELEASE]
                    [--version VERSION]
                    target [target ...]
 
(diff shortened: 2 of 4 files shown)


ehelms:
I think what you propose here is a much better solution as the default behavior of `--commit` and would ask that you just make it do so. Creating a giant blob commit when it's multiple packages is harder to manage, and doing a branch but with a commit per package sounds way better.

obal/data/playbooks/update/metadata.obal.yaml
```diff
@@ -25,5 +25,5 @@ variables:
   changelog:
     help: Optionally set the changelog. When unspecified, an entry will be generated
   commit:
-    help: When true, creates a git branch and commits the update changes to it.
+    help: When true,  commits the update changes to it into the current branch.
```
ehelms, line 28:
```suggestion
    help: When true, commits the update changes to it into the current branch.
```

obal/data/roles/git_branch_and_commit/tasks/main.yml
```diff
@@ -6,21 +6,6 @@
...
-  changed_when: false
-  register: git_branch
-  run_once: true
-
-- name: 'Create git branch'
-  command: "git checkout -b {{ git_branch.stdout }}-update-{{ inventory_hostname }}-{{ updated_version.stdout }}"
-  args:
-    chdir: "{{ inventory_dir }}"
-  run_once: true
-  changed_when: true
```
ehelms, line 22: What do you think about moving this idea to the playbook (https://github.com/theforeman/obal/blob/master/obal/data/playbooks/update/update.yaml) and only performing it once? For single packages, I find it very useful to:

```
git checkout rpm/develop
git reset origin/rpm/develop
obal update <package> --commit
git push ehelms <current branch>
```

pcreech:
@ehelms @Odilhao ping, it's been sitting a bit with no movement.

Odilhao:
I've been using this since Pulpcore 3.21 branching, it worked as intended

ehelms:
The thing missing for me here is that it removes the auto-creation of a branch for me which I have used a lot. I was hoping with this change you could keep that functionality but put it behind a parameter, e.g. `--branch`

tests/fixtures/help/update.txt
```diff
@@ -15,16 +16,21 @@ To update a package to a new version and create a git branch for the change from
...
   obal update mypackage
 
 positional arguments:
   target                the target to execute the action against
 
-optional arguments:
+options:
   -h, --help            show this help message and exit
   -v, --verbose         verbose output
+  --branch              When true, creates a git branch and commits the update changes to it.
```
ehelms, line 33:
```suggestion
  --branch              When true, creates a git branch.
```

`--commit` does the second half?

Odilhao: No --commit only commits the package update, it does not create a new branch it can be used in a loop and it will only generate news commits and not new branches

ehelms: Ohh that's a bit confusing, I would expect:

 * branch to create me a branch
 * commit to create a commit
 * branch + commit to create a new branch and commit

Odilhao: Yeah that's confusing, and I was confused. I'll refactor the code later today.

Odilhao:
After 2 years and a bunch of automation in place, I think we can drop this.

PR #328, opened 2023-01-16
KojiCommandError uses command and message, not cmd and output

evgeni:


evgeni:
CI fixes in https://github.com/theforeman/obal/pull/329

ehelms (approved):

PR #329, opened 2023-01-16
CI fixes

evgeni:


.github/workflows/main.yml
```diff
@@ -50,7 +53,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.7'
+          python-version: '3.9'
```
ekohl, line 56: Any reason you pick 3.9 here? I'd expect the latest Python version if possible.

ekohl (approved):

PR #330, opened 2023-01-16
better parse rpmbuild output

evgeni:
recent rpmbuild versions will output a lot of data on stdout and stderr, which can confuse our parsing. ignoring stderr doesn't completely help, so instead lets find the exact line saying "Wrote: some.src.rpm" and take that

### DIFF
--- a/obal/data/modules/srpm.py
+++ b/obal/data/modules/srpm.py
@@ -82,6 +82,43 @@ def fetch_remote_sources(source_location, source_system, sources_dir):
                 zip_file.extract(zip_info, sources_dir)
 
 
+def build_srpm(module, package, base_dir, sources_dir, scl=None):
+    """
+    Build the SRPM and return the path to the result
+    """
+    build_dir = os.path.join(base_dir, 'BUILD')
+    os.mkdir(build_dir)
+
+    command = ['rpmbuild', '-bs']
+    command += ['--define', '_topdir %s' % base_dir]
+    command += ['--define', '_sourcedir %s' % sources_dir]
+    command += ['--define', '_builddir %s' % build_dir]
+    command += ['--define', '_srcrpmdir %s' % base_dir]
+    command += ['--define', '_rpmdir %s' % base_dir]
+    command += ['--undefine', 'dist']
+
+    if scl:
+        command += ['--define', 'scl %s' %  scl]
+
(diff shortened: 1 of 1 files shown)


obal/data/modules/srpm.py
```diff
@@ -137,11 +137,16 @@ def main():
...
         except subprocess.CalledProcessError as error:
             module.fail_json(msg='Failed to srpm build', command=' '.join(command), output=error.output)
 
-        result = result.split("Wrote: ")[-1].rstrip()
+        result = None
+        for line in command_output.splitlines():
+            if line.startswith("Wrote: "):
+                result = line.split("Wrote: ")[-1].rstrip()
+        if not result:
+            module.fail_json(msg='Failed to find built SRPM', command=' '.join(command))
```
evgeni, line 149: pylint complains my code is too complicated :roll_eyes:

obal/data/modules/srpm.py
```diff
@@ -137,11 +137,16 @@ def main():
...
-            result = run_command(command)
+            command_output = run_command(command)
         except subprocess.CalledProcessError as error:
             module.fail_json(msg='Failed to srpm build', command=' '.join(command), output=error.output)
 
-        result = result.split("Wrote: ")[-1].rstrip()
+        result = None
+        for line in command_output.splitlines():
+            if line.startswith("Wrote: "):
+                result = line.split("Wrote: ")[-1].rstrip()
```
ekohl, line 147: Should it break after this?

You can then also consider the `else` statement for `for`, though I know plenty of people don't like that.

evgeni: Done. But yes, `for/else` ain't my biggest friend.

ekohl: I had thought you'd move the `module.fail_json` into `else`. Though I now realize that possibly you also wanted to capture empty strings. If you're not moving `fail_json` into `else`, I'd say that your previous version where you first declared `result = None` before the loop was better (but with `break` as an optimization).

ekohl (approved):
Looks good to me, but the build is rather red.

evgeni:
@ekohl this had more changes than just a limits bump, mind re-reviewing?

ekohl (approved):

PR #334, opened 2023-02-28
Disable pylint R1735(use-dict-literal)

ehelms:
As this rule would change our signature to AnsibleModule, I felt it was better to disable it for now rather than re-factor those.


evgeni:
It won't change the signature as such, but the dict() notation is what is used Overall in Ansible, so I agree we should keep it

PR #339, opened 2023-03-28
explicitly install dnf config-manager

evgeni:


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM quay.io/centos/centos:stream8
 
 RUN echo "tsflags=nodocs" >> /etc/yum.conf && \
```
ekohl, line 3: Unrelated, but does dnf still respect this?

Dockerfile
```diff
@@ -1,7 +1,7 @@
...
-    yum -y install git glibc-langpack-en epel-release python3-pip && \
+    yum -y install git glibc-langpack-en epel-release python3-pip 'dnf-command(config-manager)' && \
     yum clean all
```
ekohl, line 5: I wonder if it should use `dnf` instead of `yum`.

ekohl (approved):

PR #344, opened 2023-04-07
Refactor Copr support into modules

ehelms:
Start of re-factoring into modules based on our learning from Koji interaction. One aspect of this is how we define the data, while I am going to work on tests, I thinking of a data structure like this:

```
      copr_repo_name: 'foreman-nightly'
      chroots:
        - name: rhel-9-x86_64
          external_repo:
          modules:
        - name: rhel-8-x86_64
          external_repo:
          modules:
        - name: rhel-7-x86_64
          external_repo:
          modules:
```


ehelms:
Will require https://github.com/theforeman/obal/pull/345 to go in first and be rebased on how the modules work there.

ehelms:
I have updated this with a rebase onto master, and pushing up all the changes from testing builds locally.

ehelms:
I don't understand why this only fails on 3.11. The failing line:

https://github.com/theforeman/obal/pull/344/files#diff-103a9c682c1c4b5acd91a508a254c48b6712a50b2968f7b7c2d26f08869450d8R37

Where I have the mock printing out JSON like the `copr-cli` would:

https://github.com/theforeman/obal/pull/344/files#diff-158adaf76a7a5801ea8eafe90f5149f58986ed540ee684aa2ab498294476070eR196-R197

evgeni:
```console
% ./tests/fixtures/mockbin/copr-cli 
/home/evgeni/Devel/theforeman/obal/./tests/fixtures/mockbin/copr-cli:10: DeprecationWarning: 'pipes' is deprecated and slated for removal in Python 3.13
  from pipes import quote
{}
```

that ain't no valid json? :-)

ehelms:
> ```
> % ./tests/fixtures/mockbin/copr-cli 
> /home/evgeni/Devel/theforeman/obal/./tests/fixtures/mockbin/copr-cli:10: DeprecationWarning: 'pipes' is deprecated and slated for removal in Python 3.13
>   from pipes import quote
> {}
> ```
> 
> that ain't no valid json? :-)
> 
> Fix: #349

Thanks! I need to squint harder next time.

ehelms:
This is ready for review. It is helpful to review it with respect to https://github.com/theforeman/foreman-packaging/pull/9290 as that gives it some real world feel.

ehelms:
@evgeni What would you recommend we do here? Drop 2.7 support or test it via a different method?

```
  Warning: The support for python 2.7 will be removed on June 19. Related issue: https://github.com/actions/setup-python/issues/672
  Version 2.7 was not found in the local cache
  Error: The version '2.7' with architecture 'x64' was not found for Ubuntu 20.04.
```

obal/data/module_utils/copr.py
```diff
@@ -10,13 +10,16 @@ def __init__(self, message, command):
...
-def copr_cli(command, executable=None):
+def copr_cli(command, executable=None, config_file=None):
     """
     Run a copr-cli command
     """
     if executable is None:
         executable = 'copr-cli'
 
+    if config_file:
+        command = ['--config', config_file] + command
```
evgeni, line 21: I wish `corp-cli` would support setting this via an environment variable, but as it does not, this looks fine.

obal/data/modules/check_copr_build.py
```diff
@@ -0,0 +1,54 @@
...
+        full_name(user, project),
+        '--name',
+        package,
+        '--with-all-builds'
+    ]
+
+    try:
+        package_info = json.loads(copr_cli(command, config_file=config_file))
+    except CoprCliCommandError as error:
+        if "Error: No package with name {} in copr {}".format(package, project):
```
evgeni, line 41: this is `if <a string>:` which is always `true`, did you forget to match against some output?

obal/data/modules/check_copr_build.py
```diff
@@ -0,0 +1,54 @@
...
+            module.exit_json(changed=True, exists=False)
+        else:
+            module.fail_json(msg='Retrieval of package from Copr failed', command=command, output=error.message,
+                             repo_name=full_name(user, project), package=package)
+
+    builds = (build for build in package_info['builds'] if build['state'] == 'succeeded')
+    exists = any("{}-{}".format(package, build['source_package']['version']) == nvr for build in builds)
+
+    module.exit_json(changed=not exists, exists=exists)
```
evgeni, line 50: I find it weird that a `check_*` module has a `changed` flag -- it did not change anything.
And you don't use that in the playbooks later either, just `exists` -- so I'd just drop it here (and in the other `exit` calls)

obal/data/modules/check_copr_build.py
```diff
@@ -0,0 +1,54 @@
+"""
+Check if a build already exists in Copr
```
evgeni, line 2: scope creep: this could be a more generic `copr_package_info` module, where one of the results is whether it exists

but doesn't have to be in this PR, just food for thought

obal/data/modules/copr_chroot.py
```diff
@@ -27,23 +28,19 @@ def main():
...
-        command.extend(['--repos', ' '.join(external_repos)])
-
-    if buildroot_packages:
-        command.extend(['--packages', ' '.join(buildroot_packages)])
-
-    if modules:
-        command.extend(['--modules', ','.join(modules)])
+    command.extend(['--repos', ' '.join(external_repos)])
+    command.extend(['--packages', ' '.join(buildroot_packages)])
+    command.extend(['--modules', ','.join(modules)])
```
evgeni, line 40: the defaults for those params are `[]`, so if the user didn't pass anything, these lines will expand to:
```
command.extend(['--repos', ''])
command.extend(['--packages', ''])
command.extend(['--modules', ''])
```

is that what you intended here?

obal/data/roles/build_package/tasks/copr_build.yml
```diff
@@ -0,0 +1,120 @@
...
+      check_copr_build:
+        user: "{{ copr_project_user }}"
+        project: "{{ copr_project['copr_project_name'] }}"
+        nvr: "{{ package_nvr.nvr }}"
+        package: "{{ package_nvr.name }}"
+        config_file: "{{ build_package_copr_config | default(omit) }}"
+      register: build_exists
+
+- debug:
+    msg: "{{ build_exists.skipped is defined }}"
```
evgeni, line 24: what is this for?

obal/data/roles/build_package/tasks/copr_build.yml
```diff
@@ -0,0 +1,120 @@
...
+    - include_role:
+        name: copr_project
+      vars:
+        copr_project_user: "{{ copr_scratch_user }}"
+        copr_project_name: "{{ copr_scratch_project }}"
+        copr_project_chroots: "{{ copr_project['copr_project_chroots'] }}"
+        copr_project_description: "{{ copr_project['copr_project_description'] | default(omit) }}"
+        copr_project_delete_after_days: "4"
+        copr_project_unlisted_on_homepage: "{{ copr_project['copr_project_unlisted_on_homepage'] | default(omit) }}"
+        copr_project_copr_config: "{{ build_package_copr_config | default(omit) }}"
```
evgeni, line 86: you can't pass `default(omit)` to roles, only to modules.

tests/fixtures/mockbin/mockbin
```diff
@@ -152,4 +153,54 @@ elif prog in ['koji', 'brew']:
...
+    parser_build.add_argument('--modules')
+
+    parser_build = subparsers.add_parser('watch-build')
+    parser_build.add_argument('build')
+
+    args = parser.parse_args()
+
+    if args.subcommand == 'get-package':
+        json_data = json.loads('{"builds": [{"state": "succeeded", "source_package": {"version": "0.0.1"}}]}')
+        print(json.dumps(json_data, indent=2))
```
evgeni, line 200: why loads/dumps here? you could either take the string you have verbatim and print it (it is valid json), or have a "real" python dict and dumps it.

evgeni:
> @evgeni What would you recommend we do here? Drop 2.7 support or test it via a different method?
> 
> ```
>   Warning: The support for python 2.7 will be removed on June 19. Related issue: https://github.com/actions/setup-python/issues/672
>   Version 2.7 was not found in the local cache
>   Error: The version '2.7' with architecture 'x64' was not found for Ubuntu 20.04.
> ```

I'd prefer fixing CI (I started in https://github.com/theforeman/obal/pull/352), but that shouldn't be a blocker for now.

evgeni (changes requested):
I've posted a few questions inline, but only the `if <a string>:` I'd consider blocking. Feel free to merge in my absence once that one is cleaned out.

obal/data/modules/check_copr_build.py
ekohl: Rather than matching here, can we catch this in `copr_cli` and implement a specific `CoprPackageRetrievalError` exception (which subclasses `CoprCliCommandError`?

obal/data/modules/check_copr_build.py
```diff
@@ -0,0 +1,54 @@
...
+        package_info = json.loads(copr_cli(command, config_file=config_file))
+    except CoprCliCommandError as error:
+        if "Error: No package with name {} in copr {}".format(package, project):
+            module.exit_json(changed=True, exists=False)
+        else:
+            module.fail_json(msg='Retrieval of package from Copr failed', command=command, output=error.message,
+                             repo_name=full_name(user, project), package=package)
+
+    builds = (build for build in package_info['builds'] if build['state'] == 'succeeded')
+    exists = any("{}-{}".format(package, build['source_package']['version']) == nvr for build in builds)
```
ekohl, line 48: This is a lot to unpack, but I think this is equal to
```suggestion
    builds = (build for build in package_info['builds'] if build['state'] == 'succeeded')
    exists = any(builds)
```

Because a string is always true. So `package-False` will be `True`.

You could write it as
```suggestion
    exists = nvr in (build['source_package']['version'] for build in package_info['builds'] if build['state'] == 'succeeded')
```

obal/data/modules/copr_build.py
```diff
@@ -0,0 +1,61 @@
...
+            command.extend(['--chroot', chroot])
+
+    try:
+        output = copr_cli(command, config_file=config_file)
+    except CoprCliCommandError as error:
+        module.fail_json(msg='Copr build failed', command=error.command, output=error.message,
+                         repo_name=full_name(user, project), srpm=srpm)
+
+    build_urls = re.findall(r'^Build was added to.+:\n^\s(.+)', output, re.MULTILINE)
+    build_urls = [url.strip() for url in build_urls]
```
ekohl, line 53: I wonder if you can avoid the `strip()` call by enhancing the regex:
```suggestion
    build_urls = re.findall(r'^Build was added to.+:\n^\s+(.+)\s*', output, re.MULTILINE)
```

obal/data/roles/build_package/tasks/copr_build.yml
ehelms: Will this properly handle undefined variables? Or is default(None) needed?

evgeni: I am afraid no. No idea if `default(None)` or `default(~)` will work, try it :)

obal/data/modules/check_copr_build.py
ehelms: I am struggling to see how either of these two is equivalent to what I have. We need to compare whether the supplied nvr (name-version-release) is present in the set of builds that were returned for those builds marked as succeeded.

If you think this needs more breaking up of the logic, I could do this:

```
    successful_builds = (build for build in package_info['builds'] if build['state'] == 'succeeded')
    successful_nvrs = ("{}-{}".format(package, build['source_package']['version']) for build in builds)
    exists = nvr in successful_nvrs
```

ekohl: Ah, now I see it and I was misreading it. Your logic is sound, but I do think your proposed splitting makes sense. That's much easier to read.

Also brings me to another item: what about epochs? nvr doesn't include the epoch, so should we compare nevr?

ehelms: Epoch is a good question, I'm so used to Koji not reporting it. Looks like Copr does:

```
            "source_package": {
                "name": "rubygem-sinatra",
                "url": "https://download.copr.fedorainfracloud.org/results/@theforeman/foreman-nightly-scratch-789f37ac-5f59-5dc7-9412-b1357210f73f/srpm-builds/06166363/rubygem-sinatra-2.2.4-1.src.rpm",
                "version": "1:2.2.4-1"
            },
```

I will update the code to take Epoch into consideration.

ehelms:
The Python 2.7 tests are currently known to be broken, attempting to be fixed over here -- https://github.com/theforeman/obal/pull/352

PR #345, opened 2023-04-11
Add ability to manage a Copr project

ehelms:


obal/data/playbooks/copr-project/copr-project.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Configure Copr projects
+  hosts:
+    - copr_projects
+    - packages
```
ehelms, line 5: Hoping to avoid having to set packages here via https://github.com/theforeman/obsah/pull/25

ehelms:
This can be reviewed for basic structure and usage. I am working on adding tests.

ehelms:
The alternative approach could be to define the projects as well known variables and use a similar method to verify-koji-tag to act on the variables, using a module to do the lookup.

evgeni (commented):
Two questions:
1. this assumes `copr-cli` authentication is handled by some other means (e.g. Kerberos or `~/.config/copr`), right? (that's the same as we have with Koji, just making sure I understand)
2. did you look into using the `copr` stuff as a Python module instead of shelling out to the cli? I wouldn't block on that, but have a slight preference to the module ;)

That said, I think the existing stuff looks like it would work and be a good base

obal/data/modules/copr_project.py
```diff
@@ -0,0 +1,59 @@
...
+    delete_after_days = module.params['delete_after_days']
+
+    if not description:
+        description = project
+
+    command = [
+        'create',
+        full_name(user, project),
+        '--description',
+        "%s" % description
```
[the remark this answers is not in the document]
ehelms: :facepalm:

obal/data/modules/copr_chroot.py
```diff
@@ -0,0 +1,54 @@
...
+    buildroot_packages = module.params['buildroot_packages']
+    modules = module.params['modules']
+
+    command = [
+        'edit-chroot',
+        "%s/%s" % (full_name(user, project), chroot)
+    ]
+
+    if external_repos:
+        command.extend(['--repos', ' '.join(external_repos)])
```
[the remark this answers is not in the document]
ehelms: I was thinking: why why why

tests/test_functional.py
```diff
@@ -801,3 +801,28 @@ def test_obal_verify_tag_all():
...
+
+@obal_cli_test(repotype='copr')
+def test_copr_project_many_chroots():
+    assert_obal_success(['copr-project', 'client'])
+
+    expected_log = [
+        "copr-cli create example/foreman-client --description foreman-client --chroot rhel-9-x86_64 --chroot rhel-8-x86_64 --chroot rhel-7-x86_64 --unlisted-on-hp on",  # noqa: E501
+        "copr-cli edit-chroot example/foreman-client/rhel-9-x86_64",
+        "copr-cli edit-chroot example/foreman-client/rhel-8-x86_64",
+        "copr-cli edit-chroot example/foreman-client/rhel-7-x86_64"
```
evgeni, line 825: it's a bit weird that there are *no* edit commands at the end, but it's a valid (even if unnecessary) invocation.

IMHO something we should optimize, but not blocking

evgeni (changes requested):

tests/test_functional.py
ehelms: I do not follow what you mean here.

evgeni: `copr-cli edit-chroot example/foreman-client/rhel-7-x86_64` is a noop as it does not pass any options to change (as there are no custom settings defined for that chroot).

I think it can be optimized away, but not necessarily today.

obal/data/modules/copr_project.py
```diff
@@ -0,0 +1,59 @@
...
+    chroots = module.params['chroots']
+    description = module.params['description']
+    unlisted_on_homepage = module.params['unlisted_on_homepage']
+    delete_after_days = module.params['delete_after_days']
+
+    if not description:
+        description = project
+
+    command = [
+        'create',
```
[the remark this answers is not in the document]
ehelms: Take a look at https://github.com/theforeman/obal/pull/344/files#diff-03c091f5d96839468fd435070c76a74c346a0ba1c6f5723d51b4a10c8cb1621a

I was debating how to handle the forking for scratch builds. And this is closely related I think in approach. I considered if I should do a full-scale project module that manages it in all facets or, for the sake of complexity, keeping things separated a bit. Would you do create, edit and fork all in one?

tests/test_functional.py
ehelms: Oh I see your point: not only is this is a bad test example, it actually highlights an inefficiency if there is no edits to perform.

obal/data/modules/copr_project.py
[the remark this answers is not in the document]
evgeni: Heh, when writing the above comment, I had started another paragraph with "I guess it depends on the workflow we want to implement", but then scraped it…

The way I see this PR is that it mimics what we do today with a combo of [`external_repos.yml` in `foreman-infra`](https://github.com/theforeman/foreman-infra/blob/master/koji/playbooks/external_repos.yml) and [`tool-belt koji`](https://github.com/theforeman/tool_belt/blob/master/lib/tool_belt/koji_tag.rb) aka "manage the base we are using for building". This (for me) includes the ability to change an external repo of a chroot *after* the project has been created (possible with your code, as that is part of the `edit-chroot` code¹) or change the chroots of a project (to add el9, for example, not possible with the current code)

¹: while the edit-chroot code exists and would work, actually executing it would not work, as the module first would try a `copr-cli create` and fail :(

So, TL;DR: no. I think one module should do create/edit (aka: "manage") and another should do fork (aka: "create another one based on a template")

ehelms:
I've addressed all but the edit-config comment. Additionally, I want to note the format in the manifest, here is an additional real world example of this format: https://github.com/theforeman/foreman-packaging/pull/9290/files

The alternative approach would be how we did Koji tags, where they were treated as variables (e.g. https://github.com/theforeman/foreman-packaging/pull/9290/files#diff-8ad9be659ee8845f6bafa3ccc4f6c016ed683ebb9fb192ae9d1438613a8206fcR7) instead.

obal/data/modules/copr_project.py
```diff
@@ -0,0 +1,79 @@
...
+        'list',
+        user
+    ]
+
+    try:
+        project_list = copr_cli(command)
+    except CoprCliCommandError as error:
+        module.fail_json(msg='Copr project listing failed', command=' '.join(error.command), output=error.message)
+
+    return "Name: {}".format(project) in project_list
```
evgeni, line 23: This will be `True` if there is a project called `foreman-nightly` but the value of the `project` variable is just `foreman`.

I don't think it's a real-world issue (for our usage), as we have rather unique names for all our projects, but still IMHO something we should cleanup (at some point)

obal/data/modules/copr_project.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Create a Project in Copr
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            user=dict(type='str', required=True),
+            project=dict(type='str', required=True),
+            chroots=dict(type='list', required=True),
+            description=dict(type='str', required=False),
+            unlisted_on_homepage=dict(type='bool', required=False, default=True),
```
evgeni, line 35: Why is the default here `True`?

obal/data/modules/copr_project.py
```diff
@@ -0,0 +1,79 @@
...
+            delete_after_days=dict(type='str', required=False, default=None),
```
evgeni, line 36:
```suggestion
            delete_after_days=dict(type='str', required=False),
```

There is no need to set `default=None` if there is no default :)

obal/data/modules/copr_chroot.py
```diff
@@ -0,0 +1,54 @@
...
+    Manage a chroot in a Copr project
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            user=dict(type='str', required=True),
+            project=dict(type='str', required=True),
+            chroot=dict(type='str', required=True),
+            external_repos=dict(type='list', required=False, default=[]),
+            buildroot_packages=dict(type='list', required=False, default=[]),
+            modules=dict(type='list', required=False, default=[]),
```
evgeni, line 20: IMHO no need to set the default to `[]`

evgeni (approved):
couple of nitpicks, but feel free to ignore them for now

obal/data/modules/copr_project.py
ehelms: I could opt to do this instead with a regex so that a line terminator is included?

obal/data/modules/copr_project.py
ehelms: Originally the design included using this for forking, which I did not want forks to be taking up space, but now that we have settled on a dedicated fork module, I think it's safe for this to be defaulted to False.

obal/data/modules/copr_project.py
evgeni: Or you split the lines (`.splitlines()`) and strip them like this:
https://github.com/theforeman/obal/blob/60e8b53f1993e961f9f26a922633194648a0184a/tests/fixtures/mockbin/mockbin#L43

Then you can use the `in` on the resulting list?

Either way works

obal/data/modules/copr_project.py
evgeni: :+1:

obal/data/module_utils/copr.py
```diff
@@ -0,0 +1,29 @@
...
+
+def copr_cli(command, executable=None):
+    """
+    Run a copr-cli command
+    """
+    if executable is None:
+        executable = 'copr-cli'
+
+    try:
+        return check_output([executable] + command, universal_newlines=True)
```
evgeni, line 21: Check output only captures stdout by default.

You gotta capture stderr too, like we do in
https://github.com/theforeman/obal/blob/60e8b53f1993e961f9f26a922633194648a0184a/obal/data/module_utils/obal.py#L112

See also https://docs.python.org/3/library/subprocess.html#subprocess.check_output

ehelms:
I captured the stdout / stderr in an issue to be discussed and addressed separately as a follow up (https://github.com/theforeman/obal/issues/347). I am going to merge this as is to move forward.

obal/data/module_utils/copr.py
ehelms: Ha! What timing, I updated this.

PR #346, opened 2023-05-02
use `develop` as the default branch in tests

evgeni:
`git init` defaults to the value of `init.defaultBranch` to create new repositories, but that can vary between installations and users. excplicitly call it with `--initial-branch` to avoid that.

also update the default branch in the code to `develop` which is the default across Foreman projects

### DIFF
--- a/obal/data/roles/setup_sources/tasks/git.yml
+++ b/obal/data/roles/setup_sources/tasks/git.yml
@@ -7,7 +7,7 @@
   git:
     repo: "{{ setup_sources_git }}"
     dest: "{{ setup_sources_git_checkout_dir }}"
-    version: "{{ git_branch | default('master') }}"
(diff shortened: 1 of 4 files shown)


obal/data/roles/setup_sources/tasks/git.yml
```diff
@@ -7,7 +7,7 @@
   git:
     repo: "{{ setup_sources_git }}"
     dest: "{{ setup_sources_git_checkout_dir }}"
-    version: "{{ git_branch | default('master') }}"
+    version: "{{ git_branch | default('develop') }}"
```
ekohl, line 10: Would it be better to omit if unset? I think git then just uses whatever the remote indicates as default.

evgeni: indeed, in this case it's smarter to omit!

PR #352, opened 2023-07-11
move python 2.7 tests onto a container

evgeni:


ekohl (commented):
To drop Python 2.7 we need to make sure our CI runs on Python 3+. I think that means updating our nodes to EL8 or EL9, or using python3 explicitly. Can't we just ensure python3 is present on our nodes and change this line: https://github.com/theforeman/jenkins-jobs/blob/cb4dafc77d2f24893a1ff2fe786935e1d1c86ad7/theforeman.org/pipelines/lib/obal.groovy#L10

And then tell downstream to do the same.

ekohl:
https://github.com/theforeman/jenkins-jobs/pull/345

evgeni:
I've pulled in #360 which should make this one finally green.

evgeni:
:green_apple: 

one big downside of the "run inside container" approach is: there is no caching happening, which adds quite some time, plus the test themself seem to be a tad slower, but I prefer this over accidental breakage

ehelms (approved):

PR #353, opened 2023-07-12
Attempt to fix broken tests

ehelms:
There is odd behaviour and failure for tests with Python 3.10+:

```
PLAY [Release packages] ********************************************************

TASK [setup_workspace : Define tmp directory] **********************************
ok: [package-with-existing-build]

TASK [setup_workspace : Ensure .tmp directory] *********************************
ok: [package-with-existing-build]

TASK [git_annex_setup : Call git annex init] ***********************************
ok: [package-with-existing-build]

TASK [setup_sources : Set package_dir] *****************************************
ok: [package-with-existing-build]

TASK [setup_sources : Setup sources from specfile] *****************************
included: /home/ehelms/workspace/upstream/packaging/obal/.tox/py311/lib/python3.11/site-packages/obal/data/roles/setup_sources/tasks/specfile.yml for package-with-existing-build

TASK [setup_sources : Extract Sources from spec] *******************************
fatal: [package-with-existing-build]: FAILED! => 
  msg: |-
    The task includes an option with an undefined variable. The error was: 'spec_file_path' is undefined. 'spec_file_path' is undefined
  
    The error appears to be in '/home/ehelms/workspace/upstream/packaging/obal/.tox/py311/lib/python3.11/site-packages/obal/data/roles/setup_sources/tasks/specfile.yml': line 2, column 3, but may
    be elsewhere in the file depending on the exact syntax problem.
  
    The offending line appears to be:
  
    ---
    - name: 'Extract Sources from spec'
      ^ here
```

Where the `spec_file` role should be getting called as it's listed as a dependency of the `setup_sources` role. I am testing out a hard coded role import to see if it fixes things.

### DIFF
--- a/obal/data/roles/setup_sources/tasks/specfile.yml
+++ b/obal/data/roles/setup_sources/tasks/specfile.yml
@@ -1,4 +1,7 @@
 ---
+- include_role:
+    name: spec_file
+
 - name: 'Extract Sources from spec'
   shell: "spectool --list-files --all {{ spec_file_path }} | awk '{print $2}'"
   register: setup_sources_sources


evgeni:
this makes no sense whatsoever :exploding_head:

evgeni:
I think the real problem is the ansible version used.

Python 3.8 uses: ansible-6.7.0 ansible-core-2.13.10 
Python 3.9+ uses: ansible-8.1.0 ansible-core-2.15.1

And something got broken inbetween.

My laptop has ansible-core-2.14.6-1.fc38.noarch ansible-7.6.0-1.fc38.noarch

evgeni:
Right, I can repro this locally with tox, which fetches "latest" deps. Once I downgrade to ansible-7.7.0 ansible-core-2.14.7, tests pass again. and keeping ansible at 7.7.0 and updating core to 2.15.1 (which pip does, yelling at me that this is incompatible), breaks the tests again.

so what changed in core between 2.14 and 2.15?!

evgeni:
So here is an alternative fix for the issue: https://github.com/theforeman/obal/pull/354

But I am still puzzled *why* it think that role was already included and is not required anymore (it clearly was not, as otherwise the variable would be defined, huh)

ehelms:
Thanks for the fix in #354 !

PR #357, opened 2023-07-31
Fix SCL leaking into Copr SRPM build

ehelms:
The change to using the srpm_build role caused the scl field to become False which was interpreted as a string creating incorrect SRPM names. For example, False-nodejs-sass.

### DIFF
--- a/obal/data/roles/build_srpm/tasks/main.yaml
+++ b/obal/data/roles/build_srpm/tasks/main.yaml
@@ -3,7 +3,7 @@
   srpm:
     package: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
     output: "{{ build_srpm_output_dir }}"
-    scl: "{{ scl | default(omit) }}"
+    scl: "{{ build_srpm_scl | default(omit) }}"
     source_location: "{{ source_location | default(omit) }}"
     source_system: "{{ source_system | default(omit) }}"
   register: srpm_build


ekohl (commented):
We already omit it if it's missing:

https://github.com/theforeman/obal/blob/b7d8f1aeb856ed714181e622eb10f156c7c77224/obal/data/roles/build_srpm/tasks/main.yaml#L6

We also have this code:

https://github.com/theforeman/obal/blob/b7d8f1aeb856ed714181e622eb10f156c7c77224/obal/data/modules/srpm.py#L100-L101

I get the impression a literal string `False` is passed somehow. Perhaps it should really be `false` since that's how YAML encodes booleans.

obal/data/roles/build_srpm/tasks/main.yaml
```diff
@@ -3,7 +3,7 @@
   srpm:
     package: "{{ inventory_dir }}/{{ package_base_dir }}/{{ inventory_hostname }}"
     output: "{{ build_srpm_output_dir }}"
-    scl: "{{ scl | default(omit) }}"
+    scl: "{{ build_srpm_scl | default(omit) }}"
```
ehelms, line 6: Switching the namespaced parameter also seems to have properly fixed it.

ekohl (approved):
I like this better :)

evgeni:
And now it leaks `false` into koji builds?!

http://koji.katello.org/koji/taskinfo?taskID=562759

PR #358, opened 2023-08-01
Add comps file support for Copr chroots

ehelms:


obal/data/playbooks/copr-project/metadata.obal.yaml
```diff
@@ -5,3 +5,6 @@ variables:
   copr_project_user:
     parameter: --user
     help: Copr user to use instead of the default value
+  copr_project_copr_config:
+    parameter: --copr-config
+    help: Path to a Copr config file
```
evgeni, line 10: commit message wise, this seems not to belong here?

on the other hand, the playbook and module already do use `copr_project_copr_config`, so seems fine

evgeni (approved):

PR #360, opened 2023-08-28
Fix error output from CalledProcessError

ehelms:
Fixes:

 AttributeError: 'CalledProcessError' object has no attribute 'stdout'[0m


evgeni:
`.stdout`/`.stderr` exist since Python 3.5…, and using `.output` (which is the same as `.stdout`) would be losing stderr things (at least on 3.5+)?

evgeni:
OTOH, we use `.output` everywhere else, so :woman_shrugging:

PR #362, opened 2023-08-30
Allow running repoclosure based on a URL and dist

ehelms:
Example:

```
obal-source repoclosure rubygem-pulp_file_client --check https://download.copr.fedorainfracloud.org/results/@theforeman/foreman-katello-nightly-staging-scratch-0b6bfd0a-f472-5d75-9280-3ae62f547728/rhel-8-x86_64/ --dist el8
```

This will allow us to stop downloading RPMs to create a local repository from and instead use the repositories created by the scratch repos from Copr to do repoclosure.

### DIFF
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -657,6 +657,22 @@ def test_obal_repoclosure_katello_with_downloaded_rpms():
     ]
     assert_in_mockbin_log(expected_log)
 
+@obal_cli_test(repotype='upstream')
+def test_obal_repoclosure_with_check_repo():
+    assert_obal_success(['repoclosure', 'hello', '--check', 'https://test.example.com/myrepo', '--dist', 'rhel7'])
+
+    expected_log = [
+        "dnf repoclosure",
+        "--newest",
+        "--refresh",
+        "--config",
+        "repoclosure/yum.conf",
+        "--check repo0",
+        "--repofrompath repo0,https://test.example.com/myrepo",
+        "--repo el7-base"
+    ]
+
+    assert_in_mockbin_log(expected_log)
 
 @obal_cli_test(repotype='copr')
 def test_obal_scratch_copr_hello_nowait():
--- a/obal/data/playbooks/repoclosure/metadata.obal.yaml
+++ b/obal/data/playbooks/repoclosure/metadata.obal.yaml
@@ -3,3 +3,11 @@ help: |
   Run repoclosure for a repository
 
   Repoclosure ensures that all dependencies are met and all packages are installable.
+variables:
+  repoclosure_check_repos:
+    store: append
+    parameter: --check
+    help: Link to a repository to target for repoclosure
+  repoclosure_target_dist:
+    parameter: --dist
+    help: Target dist to test against (e.g. el8)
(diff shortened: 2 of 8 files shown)


ehelms:
@evgeni Take a look when you can

evgeni:
So I am trying to wrap my head around "why do we need to introduce new args".

The URL makes sense -- it changes on every scratch build, and that is the one we wanna check.
But the dist?

tests/test_functional.py
```diff
@@ -655,6 +655,23 @@ def test_obal_repoclosure_katello_with_downloaded_rpms():
...
+    assert_obal_success(['repoclosure', 'hello', '--check', 'https://test.example.com/myrepo', '--dist', 'el7'])
+
+    expected_log = [
+        "dnf repoclosure",
+        "--config",
+        "repoclosure/yum.conf",
+        "--newest",
+        "--check downloaded_rpms",
+        "--check check_repo",
+        "--repofrompath=check_repo,https://test.example.com/myrepo",
```
evgeni, line 669: this looks weird where we tell to `--check` two repos, but only really define one of them. I know it works, but still.

Is there ever a case where we want both the remote repo and "downloaded rpms" be checked together?
Otherwise we could call both repos something like `check_repo` and only pass `--check` once?

ehelms:
> So I am trying to wrap my head around "why do we need to introduce new args".
> 
> The URL makes sense -- it changes on every scratch build, and that is the one we wanna check. But the dist?

The dist is what tells our code which set of lookaside repositories to use -- https://github.com/theforeman/foreman-packaging/blob/rpm/develop/package_manifest.yaml#L155-L156

tests/test_functional.py
ehelms: It is sort of ugly and tricky due to how much if/else checking and construction has to happen. For example, constructing this variable (https://github.com/theforeman/obal/pull/362/files#diff-1f4bf62d8dfa1c838d5fae334e87225b942dc26f92086eb2ab754cbe77237a24R4). Honestly, I feel like the whole role needs an overhaul and introduction of a module to handle things cleanly and do the logic more cleanly inside python code. 

My initial goal with this support was to not break the Koji method while providing a fix for the Copr method.

ehelms: In addition, this gets a big ugly too because in CI I am going to have to write the copr repository out to a file from the build step that can then be used as input to the repoclosure step.

An all-in-one `obal scratch` is nice for users, but it creates a bit of yuck in CI where we need to pass data around from invocation to invocation.

Status output has a similar issue, having to print out a yaml file so that the status generation can then slurp it in (https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/lib/copr.groovy#L1).

ehelms: Ehh, I decided to go ahead and write a module and then I can re-factor on top of that: https://github.com/theforeman/obal/pull/364

ehelms:
This has been completely re-worked and re-designed using the module. Expected usage:

```
obal-source repoclosure rubygem-pulp_file_client --check https://download.copr.fedorainfracloud.org/results/@theforeman/foreman-katello-nightly-staging-scratch-0b6bfd0a-f472-5d75-9280-3ae62f547728/rhel-8-x86_64/ --dist el8
```

obal/data/roles/repoclosure/tasks/rpms_repoclosure.yml
```diff
@@ -0,0 +1,64 @@
+- block:
```
evgeni, line 1: what is this file for?

evgeni (approved):
I think you got a stray file in here, but otherwise LGTM

PR #364, opened 2023-09-07
Refactor repoclosure into a module

ehelms:


obal/data/modules/repoclosure.py
```diff
@@ -0,0 +1,55 @@
...
+        '--config',
+        config
+    ]
+
+    for name in check:
+        command.extend(['--check', name])
+
+    for repo in additional_repos:
+        name, url = repo.split(',')
+        command.extend(['--repofrompath', f"{name},{url}"])
```
evgeni, line 41: this implies that we call

```
repoclosure:
  additional_repos:
    - something,https://somewhere
    - else,https://somewhereelse
```

right? we could also make that param a list of dicts, which then will make the call:
```
repoclosure:
  additional_repos
    - name: something
      url: https://…
    - name: else
      url: https://…
```

The later is imho more readable (in the playbook), but I have no idea who/where the input is generated from right now.

obal/data/modules/repoclosure.py
```diff
@@ -0,0 +1,55 @@
...
+
+    for repo in lookaside:
+        command.extend(['--repo', repo])
+
+    try:
+        output = check_output(command, universal_newlines=True, stderr=STDOUT)
+    except CalledProcessError as error:
+        module.fail_json(msg='Repoclosure failed', command=' '.join(command), output=error.output)
+
+    module.exit_json(changed=True, output=output)
```
evgeni, line 51: I'd argue that this is never `changed`? I mean, the call did not perform any changes, right?

obal/data/modules/repoclosure.py
ehelms: I was torn between these two worlds. One being the follow the way repoclosure defines additional repositories and make it easier to do so via a command line parameter vs. the clean yaml method of using a dictionary.

obal/data/roles/repoclosure/tasks/repoclosure.yml
```diff
@@ -27,23 +27,15 @@
...
-            --newest
-            --check {{ repoclosure_target_repos_all | join(' --check ') }}
-            {{ '--repofrompath=downloaded_rpms,./downloaded_rpms/' + dist if repoclosure_use_downloaded_rpms else '' }}
-            {{ additional_repos | join(' ') }}
-            {{ '--repo ' if repoclosure_lookaside_repos else '' }}
-            {{ repoclosure_lookaside_repos[dist] | join(' --repo ') }}
+      repoclosure:
+        config: "{{ temp_directory.path }}/{{ config_filename }}"
+        check: "{{ repoclosure_target_repos_all }}"
+        additional_repos: "{{ additional_repos + ['--repofrompath=downloaded_rpms,./downloaded_rpms/' + dist] }}"
```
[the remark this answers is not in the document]
ehelms: > Edit: ah! `assert_in_mockbin_log` only checks for those entries being "in", and `--repofrompath=downloaded_rpms,./downloaded_rpms/` is of course _in_ `--repofrompath=--repofrompath=downloaded_rpms,./downloaded_rpms/`

Well who wrote it that way! :rofl:  I'll be honest, I didn't think this should work, and then the tests passed and my local call passed so I was like it must work.

obal/data/modules/repoclosure.py
evgeni: I think the way `dnf` expects its CLI parameters is an implementation detail that should not leak to the user of the module.
(and you can print whatever is in `command` in the error for reproducing locally).

That said, this is not blocking.

tests/fixtures/testrepo/upstream/package_manifest.yaml
```diff
@@ -33,6 +33,9 @@ repoclosures:
   hosts:
     core-repoclosure:
       repoclosure_target_dist: rhel7
+      repoclosure_target_repos:
+        rhel7:
+          - el7-katello
```
evgeni, line 38: why was this addition needed?

tests/test_functional.py
```diff
@@ -607,10 +607,10 @@ def test_obal_repoclosure():
 
     expected_log = [
         "dnf repoclosure",
+        "--newest",
+        "--refresh",
         "--config",
         "repoclosure/yum.conf",
-        "--newest",
-        "--check downloaded_rpms",
```
evgeni, line 613: it looks like now it did not call any `--check` repos, which is probably wrong?

evgeni: or what is that test testing? there is no `downloaded_rpms` fake created (like in L622)

tests/test_functional.py
```diff
@@ -644,12 +645,14 @@ def test_obal_repoclosure_katello_with_downloaded_rpms():
...
+        "--newest",
+        "--refresh",
         "--config",
         "repoclosure/yum.conf",
-        "--newest",
         "--check el7-katello",
         "--check downloaded_rpms",
-        "--repofrompath=downloaded_rpms,./downloaded_rpms/rhel7",
+        "--repofrompath downloaded_rpms,{pwd}/downloaded_rpms/rhel7",
+        "downloaded_rpms/rhel7",
```
evgeni, line 655: that second `downloaded_rpms` is probably an oversight during editing?

tests/test_functional.py
ehelms: There is something weird going on with this test, as it does not appear to be testing for *anything*. I can change the expected log to "ahdslkjsdf" and it still passes. I am in investigating.

evgeni (approved):

PR #370, opened 2023-10-01
repoclosure_check_repos is a list, so let's pass it as one

evgeni:
`loop` is picky when you pass a list of lists


ehelms (approved):

obal/data/playbooks/repoclosure/metadata.obal.yaml
```diff
@@ -5,7 +5,7 @@ help: |
   Repoclosure ensures that all dependencies are met and all packages are installable.
 variables:
   repoclosure_check_repos:
-    store: append
+    action: append
```
evgeni, line 8: well, this explains why it never failed in CI before ;-)

PR #372, opened 2023-10-13
Add ability to fork instead of create a project

ehelms:


obal/data/roles/copr_project/tasks/main.yaml
```diff
@@ -1,25 +1,6 @@
...
-    chroot: "{{ chroot.name }}"
-    external_repos: "{{ chroot.external_repos | default(omit) }}"
-    buildroot_packages: "{{ chroot.buildroot_packages | default(omit) }}"
-    modules: "{{ chroot.modules | default(omit) }}"
-    config_file: "{{ copr_project_copr_config | default(omit) }}"
-    comps_file: "{{ chroot.comps_file | default(omit) }}"
-  loop: "{{ copr_project_chroots }}"
-  loop_control:
-    loop_var: chroot
+- when: not copr_project_fork_from
```
evgeni, line 5: I think you gotta add `copr_project_fork_from: False` to the defaults of that role

PR #377, opened 2023-12-11
Add support for rpmlint folder in the root of inventory_dir

Odilhao:
This PRs adds support for one folder called rpmlint inside of the inventory_dir.

With Pulp packaging going to Python 3.11, rpmlint is about to break when the Jenkins nodes start to use one ansible version that requires python 3.11 and install python3.11-libs, it will start to complain about the
  dist-utls deprecation in python 3.12+.

### DIFF
--- a/obal/data/playbooks/lint/lint.yaml
+++ b/obal/data/playbooks/lint/lint.yaml
@@ -7,6 +7,20 @@
   roles:
     - package_variables
   tasks:
+
+    - name: 'Scans for rpmlint file'
+      ansible.builtin.find:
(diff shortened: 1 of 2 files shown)


evgeni:
What's the reason to do it like this vs setting rpmlint_rc_file in the manifest?

I would expect this to break when someone wants to set rc file in the manifest, but there is one in the folder?

Odilhao:
Sometimes I forget that we can use the manifest as one Ansible Inventory, I need to buy one rubber duck

evgeni:
E-v6TS_XEAAEvRC

PR #378, opened 2023-12-20
make copr forking safe

evgeni:


ehelms (approved):

evgeni:
tested in production on `plugins-3.9-staging`

PR #380, opened 2023-12-23
Drop tito integration

### LINKED ISSUE #294: Drop tito as a dependency
Tito serves as middleman for operations to Koji. Projects can end up duplicating information in tito configuration that is present in the package manifest. This issue aims to replace all the functionality tito handled with native functionality to drop tito and drive all configuration through the package manifest as the source of truth.

Target items:

 * Running Koji commands directly to perform scratch builds
 * Running Koji commands directly to perform releases
 * Handling dist-git commit

ekohl:
We have moved away from tito and everything runs with direct koji commands now.

Fixes #294


obal/data/roles/setup_sources/tasks/git.yml
```diff
@@ -10,8 +10,3 @@
     version: "{{ git_branch | default(omit) }}"
     force: true
     depth: 1
-
-- name: 'Register source dir tito argument'
-  set_fact:
-    build_package_tito_releaser_args:
-      - 'source_dir={{ setup_sources_git_checkout_dir }}'
```
ekohl, line 17: Now that this is dropped, `setup_sources_git_checkout_dir` isn't used anywhere anymore. That means the role now just clones a git repository and then never uses it anymore. Is this role now redundant too? `setup_sources_git` was introduced in 923d72cd0c65ff652d052cb531c06909a286a0c7 but I think this is Pulp  2. https://github.com/pulp/pulp-packaging/blob/master/package_manifest.yaml is the only place I could find that used it.

evgeni:
Is that true for brew builds?

Cc @pcreech @Odilhao

Odilhao:
> Is that true for brew builds?
> 
> Cc @pcreech @Odilhao 

We need tito for brew, ww still don't have a way to use dist-git without tito, the plan is to bring this to obal next year.

ekohl:
Marking it as draft until then.

pcreech:
There is still a need, yes, for obal to use tito to interact with rpkg based tooling (rhpkg, fedpkg) to interact with distgit.  It's a very minimal segment at this point, just the "obal release" workflow.  It shouldn't be too hard to utilize these tools in this workflow.

PR #381, opened 2023-12-23
Depend on ansible-core on Python 3.6+

ekohl:
This makes packaging on Fedora easier, where the is no more ansible but only ansible-core.

I still maintain https://copr.fedorainfracloud.org/coprs/ekohl/obal but ran into problems while updating and upgrading to Fedora 39 where ansible has been dropped. I believe we don't depend on any modules outside of ansible-core and our own so it also makes installation in other places faster.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,11 @@ setup(
 
     packages=find_packages(exclude=['contrib', 'docs', 'tests']),
 
-    install_requires=['ansible >= 2.5', 'obsah >= 0.0.3'],
+    install_requires=[
+        'ansible >= 2.5; python_version < 3.6',
+        'ansible-core >= 2.11; python_version >= 3.6',
+        'obsah >= 0.0.3',
+    ],
 
     extras_require={
         'argcomplete': ['argcomplete'],


evgeni:
Obsah already has that:
https://github.com/theforeman/obsah/blob/d07740690ce3100c72fc0b0d0cc3a7c817453b79/setup.py#L68

So we can just drop the Ansible requirement all together if we do not need anything from non core.

But at the same time: Fedora still ships Ansible? https://src.fedoraproject.org/rpms/ansible

ekohl:
I knew I already wrote that code! Just couldn't find it in obal. I'd be fine with dropping it here. I don't think we need anything more specific than what obsah pulls in.

And I saw a build failure on the ansible dependency. Perhaps it no longer provides `python(ansible)` then?

evgeni:
Yeah `python(ansible)` sounds like a Python lib, which should be in `ansible-core`?

evgeni:
Drop the requirement and see if CI passes?

ekohl:
I'm not at home and don't have my laptop. Something for later

evgeni:
Same :)

ekohl:
https://github.com/theforeman/obal/pull/382

evgeni:
> https://github.com/theforeman/obal/pull/382

This was merged, which makes this PR obsolete

PR #382, opened 2023-12-28
Rely on obsah to pull in Ansible

ekohl:
obsah already pulls in either ansible or ansible-core, depending on the Python version. There's no need for obal itself to pull it in.

### DIFF
--- a/obal/data/ansible.cfg
+++ b/obal/data/ansible.cfg
@@ -3,7 +3,6 @@ inventory = package_manifest.yaml
 retry_files_enabled = False
 transport = local
 nocows = 1
-stdout_callback = yaml
 display_skipped_hosts = false
 roles_path = ./roles/
 library = ./modules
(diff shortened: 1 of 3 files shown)


evgeni:
CI says "no"

evgeni:
```
6 failed, 195 passed
```

okay, not *that* bad.

ekohl:
At least this part fails:
https://github.com/theforeman/obal/blob/3625419b8656e048a76f7ef8653b86be59ebee3f/obal/data/roles/fetch_sources/tasks/main.yml#L20-L24

I think that causes most if not all failures.

ekohl:
Packaging wise we can depend on `ansible-collection-ansible-posix` but I'm not sure how to deal with pip based installations.

evgeni:
This escalated not as much as I have feared. What are we missing? :)

ekohl:
And copy works the same here? I don't know it well enough to say but I trust you

evgeni:
It doesn't *sync*, but it does recursively copy, which should be sufficient for our usecase.

And the usage should be covered by tests, so…

ekohl:
Thanks! I've reordered the commits and changed the commit messages, but no actual code changes. I believe this is ready to go, but can't approve my own PR.

evgeni (approved):

PR #387, opened 2024-04-17
only check repoclosure for noarch and x86_64 RPMs

evgeni:
especially, don't check src, as they might have deps on development tooling, that is not present in normal lookaside repos.

this is only really neccessary for copr repos that ship both src and binary in the same repo.


ekohl (approved):
I wish there was a way to exclude source packages, but I guess we'll need to do a big effort for non-x86_64 (like ARM) anyway. Perhaps at least introduce it as a parameter defaulting to `x86_64`?

evgeni:
@ekohl like this?

ehelms (approved):

ekohl (approved):
Yes, like this

PR #394, opened 2024-09-25
default to el9 for mock builds

evgeni:


adamruzicka (approved):
+1, not sure about the test failures though

evgeni:
Those are unrelated.

PR #397, opened 2024-10-28
Drop wait.yml and put waiting code in each respective build type

ehelms:
The logic for how to wait on a task to build across the different build types supported has diverged to the point where a separate file is more confusing than it is helpful in understanding the code logic.


ehelms:
Failing 3.6/3.7/3.8 is expected due to pylint.

obal/data/roles/build_package/tasks/copr_build.yml
```diff
@@ -56,9 +56,22 @@
...
+    - name: Extract Copr builds
+      set_fact:
+        copr_build_ids: "{{ copr_builds['results'] | selectattr('builds', 'defined') | map(attribute='builds') | list | flatten }}"
+
+    - name: "Watch copr build(s)"
+      command:
+        "copr-cli {{ '--config ' + build_package_copr_config if build_package_copr_config is defined else '' }} watch-build {{ copr_build_ids | join(' ') }}"
+      ignore_errors: true
+      register: build_package_copr_status
+      changed_when: false
```
evgeni, line 69: you can do:
```suggestion
      changed_when: false
      vars:
        copr_build_ids: "{{ copr_builds['results'] | selectattr('builds', 'defined') | map(attribute='builds') | list | flatten }}"
```

To save the `set_fact`

tests/test_functional.py
```diff
@@ -168,10 +168,8 @@ def test_obal_scratch_with_koji_upstream_hello():
     expected_log = [
         "koji build obaltest-nightly-rhel7 /tmp/SRPMs/hello-2.10-2.src.rpm --scratch",
         "koji watch-task 1234",
-        "koji taskinfo -v 1234",
```
evgeni, line 171: on the one hand, this seems unexpected.

on the other, we don't use `koji build` anywhwere anymore, right?
it's either `copr build` or `tito`?

obal/data/roles/build_package/tasks/copr_build.yml
ehelms: I fail, you taught me this already.

evgeni (approved):

PR #404, opened 2025-02-12
Update CI: use Ubuntu 24.04, add Python 3.13, drop Python 3.6 - 3.8

evgeni:


ekohl (approved):
We no longer have any 3.6? Also not in downstream?

evgeni:
> We no longer have any 3.6? Also not in downstream?

Correct, everything runs on at least EL9 :)

Odilhao:
> > We no longer have any 3.6? Also not in downstream?
> 
> Correct, everything runs on at least EL9 :)

Kinda but I agree

PR #409, opened 2025-04-16
Set LANG to C for rpmbuild

ehelms:
The code relies on the output from the rpmbuild command and if a different language is used this code will fail to detect the SRPM path.


lubomir:
Works great! :+1:

PR #411, opened 2025-08-26
Replace pkg_resources with importlib for Python 3.12+ support

### LINKED ISSUE #405: Remove deprecated pkg_resources in order to support Python 3.12+
https://github.com/theforeman/obal/blob/c2c3d2e1fb2141eb150e7bf7c99a0ab2dcd116cd/obal/__init__.py#L13

ogajduse:
## Summary
• Replace deprecated pkg_resources with modern importlib
• Fix Python 3.12+ compatibility issues
• Update shebang from python2 to python3

Fixes #405

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ All `obal` actions should also work with plain Ansible when called like `ansible
 
 ## necessary tools
 
-- `python` (2 or 3)
+- `python` 3.9+
 - `ansible`
 
 ## Using Obal via Container


ogajduse:
`importlib.resources.files` was added in Python 3.9. If 3.9 is the lowest version we support, we should be fine. 
https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files

This patch will allow running on Python > 3.11.

docs/source/conf.py
```diff
@@ -24,10 +24,8 @@
 author = 'The Foreman'
 
 try:
-    import pkg_resources
-    package = pkg_resources.require("obal")[0]
-    version = package.version
-    release = package.version
+    from importlib import metadata
+    version = release = metadata.version('obal')
 except:
```
ekohl: Shall we scope it explicitly to `metadata.PackageNotFoundError` now?

I'd be tempted to use
```python
from importlib import metadata
try:
    version = release = metadata.version('obal')
except metadata.PackageNotFoundError:
    version = release = ''

File: obal/__init__.py [hunk 2]
```diff
@@ -46,12 +46,7 @@ def data_path():
         """
         path = os.environ.get('OBAL_DATA')
         if path is None:
-            path = pkg_resources.resource_filename(__name__, 'data')
-            if not os.path.isabs(path):
-                # this is essentially a workaround for
-                # https://github.com/pytest-dev/pytest-xdist/issues/414
-                distribution = pkg_resources.get_distribution('obal')
-                path = os.path.join(distribution.location, path)
+            path = str(resources.files(__name__) / 'data')
```
ekohl: Note to self: added in Python 3.9 (https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files). This call looks correct for both 3.9 and 3.12+ (where it was changed).

ekohl (approved):
Looks good. I don't consider the docs config to be blocking and I need this on my own dev setup so merging this.

ekohl:
I opened https://github.com/theforeman/obal/pull/417 for the rpmlint warning but that still failed so I also opened https://github.com/theforeman/obal/pull/418 to address the other errors you found and reported in #412.

ekohl (approved):
Merging despite failing CI. That's introduced by https://docs.ansible.com/projects/ansible-core/devel/porting_guides/porting_guide_core_2.19.html#example-implicit-boolean-conversion and tracked in https://github.com/theforeman/obal/issues/412.

PR #436, opened 2026-07-08
Re-add --best to repoclosure

ogajduse:
## Summary

Without `--best`, `dnf repoclosure` can silently resolve a `Requires` against an older, still-present package instead of the newest one — masking real reverse-dependency breaks at PR time instead of failing loudly.

Real-world case: theforeman/foreman-packaging#13647 bumped `rubygem-foreman-tasks` to 13.0.0 while `rubygem-foreman_ansible` still had `Requires: (rubygem(foreman-tasks) >= 10.0 with rubygem(foreman-tasks) < 13)`. Per-PR COPR repoclosure passed anyway. The break wasn't caught until staging repoclosure ran the next day, fixed in theforeman/foreman-packaging#13661 (and its DEB-side twin, theforeman/foreman-packaging#13688).

`--best` was originally dropped in #246 (2020) because it broke EL8 module repoclosure (`module_hotfixes` wasn't respected). `module_hotfixes=1` has since been added to the modular repos in `foreman-packaging`'s `repoclosure/yum.conf`, which already neutralizes that problem — so the original reason for dropping `--best` no longer applies.

## Verification

Reproduced the actual incident with synthetic RPMs mirroring the `foreman-tasks`/`foreman_ansible` versions involved, run through the real dnf4 client (CentOS Stream 9, matching this project's own `Dockerfile` — not dnf5, whose `repoclosure --best` is a no-op alias for `--newest` and would give a false read here):

- **Without `--best`:** `dnf repoclosure` exits `0` — silently passes, reproducing the real miss.
- **With `--best`:** exits `1` and correctly reports the unresolved `rubygem(foreman-tasks) < 13` dependency.

Also re-ran today's real `el8-foreman-client-nightly-staging` and `el9-foreman-plugins-nightly-staging` targets (with their full lookaside sets) with `--best` — clean, no new false positives.

`tests/test_functional.py` updated to expect `--best` in the generated `dnf repoclosure` command. Full suite: 126 passed.

## Related

- Slack discussion (@adamruzicka, @evgeni, and myself) on making repoclosure catch cross-package breakage at PR time instead of relying on the releaser to think of it.
- The complementary fix (staging repos as lookaside for per-PR checks) is already implemented in `foreman-packaging`'s `package_manifest.yaml` — no obal-side gap there.


evgeni (approved):

zjhuntin (approved):
## Review: Re-add --best to repoclosure

**Scope**: 2 files changed, +5/-0 lines | Adds `--best` flag to the `dnf repoclosure` command

---

### Summary

This PR re-adds the `--best` flag to `dnf repoclosure` that was dropped in PR #246 (2020) due to EL8 module conflicts. The original reason no longer applies since `module_hotfixes=1` was added to the modular repos in `foreman-packaging`'s repoclosure config. Without `--best`, dnf can silently resolve a `Requires` against an older package instead of the newest, masking real dependency breakage at PR time. Ready to merge.

### Verdict: APPROVE

---

### Critical Issues (must fix)

None.

### Important Issues (should fix)

None.

### Minor Suggestions (nice to have)

None — the change is exactly one line in the module and the corresponding test updates. There's nothing to simplify or improve.

zjhuntin (approved):
## Review: Re-add --best to repoclosure

**Scope**: 2 files changed, +5/-0 lines | Adds `--best` flag to the `dnf repoclosure` command

---

### Summary

This PR re-adds `--best` to the `dnf repoclosure` command in `obal/data/modules/repoclosure.py`. The flag was dropped in PR #246 (2020) because it broke EL8 module repoclosure, but that's been neutralized by `module_hotfixes=1` in foreman-packaging's `repoclosure/yum.conf`. Without `--best`, dnf can silently resolve a `Requires` against an older package instead of the newest, masking real dependency breakage — as demonstrated by the foreman-packaging#13647 incident. Ready to merge.

### Verdict: APPROVE

---

### Critical Issues (must fix)

None.

### Important Issues (should fix)

None.

### Minor Suggestions (nice to have)

None. The code change is a single line addition at the right position in the command array (after `--newest`, grouping resolution-strategy flags together). All 4 repoclosure functional tests (`test_obal_repoclosure`, `test_obal_repoclosure_with_downloaded_rpms`, `test_obal_repoclosure_katello_with_downloaded_rpms`, `test_obal_repoclosure_with_check_repo`) are updated. The `assert_in_mockbin_log` helper does per-entry substring matching so the flag ordering in the test lists is flexible.

Odilhao (approved):

PR #441, opened 2026-08-03
Remove old local (non-URL) sources during update

### LINKED ISSUE #84: obal update fails if Source0 is a local file instead of an URL

ogajduse:
## Summary

`obal update`'s "Remove old sources" task in `obal/data/playbooks/update/update.yaml`
only ran `git rm` on Source entries containing `://`, skipping local/bare-filename
sources entirely:

```yaml
- name: "Remove old sources"
  command: "git rm --ignore-unmatch {{ item | basename }}"
  ...
  when:
    - "'://' in item"
```

That guard traces back to 7be38f5 ("Only remove sources if remote"), following
b3024bb which closed #84/#89 (`obal update` crashing on non-URL `Source0`
entries when trying `git annex add`). Nothing there suggests removal of local
sources was meant to be skipped forever — the guard looks copied onto this task
by symmetry with the annex-add task, where it's genuinely required. `git rm
--ignore-unmatch` is already safe on any path (existing or not), so the guard
was never load-bearing here.

**Effect of the bug:** whenever a local (non-URL) source's filename changes —
a `%{version}` bump, or a packaging-strategy rename — the old file is never
removed and becomes a permanent orphan. Confirmed in the wild via
theforeman/foreman-packaging's `nodejs-history` and
`nodejs-compression-webpack-plugin`, where a recent npm2rpm strategy change
renamed the bundled local source from `*-registry.npmjs.org.tgz` to
`*-package-lock.json` (same `Version`, `Release` bump only) and downstream
consumers ended up with both the old and new tarballs checked in.

## Fix

Drop the `'://' in item` condition on the "Remove old sources" task. The
symmetric guard on `git annex add`/`git add` for *new* sources is untouched
and still correct.

## Test plan

- [x] Added `test_obal_update_downstream_removes_renamed_local_source`,
      exercising the same upstream-git-clone-and-copy `downstream` flow as
      the existing `test_obal_update_downstream_hello` tests, with a fixture
      package (`bar`) whose local `Source0` is renamed upstream (`Version`
      unchanged, `Release` bumped) — mirroring the real nodejs scenario.
- [x] Verified red/green manually: fails (old file survives) with the guard
      restored, passes with it removed.
- [x] Full functional suite passes under `tox -e py313` (pytest-pylint
      incompatibility with current pytest on `master` is pre-existing and
      unrelated — worked around locally by skipping that stage; not
      addressed by this PR).


zjhuntin (approved):
## Review: Remove old local (non-URL) sources during update

**Scope**: 8 files changed, +82/-2 lines | Removes a `when` guard on the "Remove old sources" task and adds a functional test

---

### Summary

This PR fixes a real bug where local (non-URL) source files were never cleaned up during `obal update` because the `'://' in item` guard on the `git rm --ignore-unmatch` task incorrectly filtered them out. The fix is a 2-line deletion. The test is thorough and mirrors the real-world scenario (nodejs bundle renames in foreman-packaging). Ready to merge.

### Verdict: APPROVE

---

### Critical Issues (must fix)

None.

### Important Issues (should fix)

None.

### Minor Suggestions (nice to have)

- **`tests/fixtures/testrepo/downstream/packages/bar/bar.spec:21`**: The `%files` section lists `bar-old-source` as a bare filename, which works for the test but would fail an actual RPM build (no `%{buildroot}` prefix path). The existing `hello` fixtures follow the same minimal pattern so this is consistent — just noting for awareness.

zjhuntin (approved):
## Review: Remove old local (non-URL) sources during update

**Scope**: 8 files changed, +82/-2 lines | Removes a `when` guard on the "Remove old sources" task and adds a functional test

---

### Summary

This PR fixes a bug where local (non-URL) source files were never cleaned up during `obal update` because the `'://' in item` guard on the "Remove old sources" task incorrectly filtered them out. The fix is a 2-line deletion — the symmetric guards on `git annex add` / `git add` for *new* sources (lines 42–51) are correctly left in place since those route URL sources through annex and local sources through regular git. The `command` module with `git rm --ignore-unmatch` is already safe on any path, so the guard was never load-bearing here. Ready to merge.

### Verdict: APPROVE

---

### Critical Issues (must fix)

None.

### Important Issues (should fix)

None.

### Minor Suggestions (nice to have)

None. The fix is minimal, the test follows established patterns (`@obal_cli_test(repotype='downstream')`, `setup_upstream`, `assert_obal_success`, file existence checks, spec content assertions), and the fixtures are well-designed — same `Version`, bumped `Release`, renamed `Source0` mirrors the real npm2rpm scenario without overcomplicating things.

PR #442, opened 2026-08-04
Scope repoclosure's persistent-repo exclude to the package's own binary names

ogajduse:
## Summary

`--newest --best` (#436) resolves `Requires` against the newest package across *all* repos
combined, while still checking every repo's own newest package in isolation. A subpackage that
self-pins via `Requires: %{name} = %{version}-%{release}` (e.g. a `-doc`/`-cockpit` sibling)
then false-positives the instant a newer build of the same SRPM lands anywhere else in the
checked set — the persistent staging repo's stale copy has nothing left to satisfy its exact
pin against.

Reproduced against the real `theforeman/foreman-packaging#13921` and `#13922` CI failures
(`rubygem-foreman_remote_execution` and `rubygem-foreman_leapp` bumps), using the live
`el9-foreman-plugins-nightly-staging` repo and the real `repoclosure/yum.conf`.

## Fix

Exclude the package under test from the persistent staging repos it's checked against, since
the PR's own freshly-built repo (`repo0`/`downloaded_rpms`) is authoritative for it. Scoped
with dnf's per-repo `--setopt=<repo>.excludepkgs=`, never applied to the fresh repo itself.

Exclude list is resolved to the package's **exact binary names** via `rpmspec`, never a
name-prefix glob — a glob would also match unrelated packages sharing a name prefix (e.g.
`foreman*` also matching `foreman-proxy`, `rubygem-hammer_cli*` also matching
`rubygem-hammer_cli_foreman`), silently masking a real cross-package regression on those
packages — exactly the failure mode `--best` (#436) exists to catch
(`theforeman/foreman-packaging#13647`/`#13661`/`#13688`). Falls back to the exact hostname
(never a glob) if the spec fails to parse; resolves to no exclusion at all for repoclosure-only
hosts with no package directory.

## Testing

- New unit tests for `repoclosure.py`'s extracted `build_command()`.
- New/updated functional tests covering: single and multiple persistent target repos, the
  `downloaded_rpms` release flow, repoclosure-only (spec-less) hosts, and a package with an
  intentionally unparseable spec (exercises the exact-hostname fallback branch).
- Full suite: 135 passed. `pylint`: 10.00/10.
- Manually re-validated against real dnf4 (matching the CI environment) with synthetic and
  live-repo reproductions of both the original false positive and the cross-package regression
  `--best` must keep catching.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/obal/data/modules/repoclosure.py
+++ b/obal/data/modules/repoclosure.py
@@ -7,24 +7,16 @@ from subprocess import check_output, CalledProcessError, STDOUT
 from ansible.module_utils.basic import AnsibleModule
 
 
-def main():
+def build_command(config, check, additional_repos=None, lookaside=None, arch=None,
+                  exclude_repos=None, exclude_packages=None):
     """
-    Run repoclosure on a set of repositories
+    Build the `dnf repoclosure` command line for the given set of repositories
     """
-    module = AnsibleModule(
-        argument_spec=dict(
-            config=dict(type='str', required=True),
-            check=dict(type='list', required=True),
-            additional_repos=dict(type='list', required=False, default=[]),
-            lookaside=dict(type='list', required=False, default=[]),
-            arch=dict(type='list', required=False, default=['noarch', 'x86_64']),
-        )
-    )
-
-    config = module.params['config']
-    check = module.params['check']
-    additional_repos = module.params['additional_repos']
-    lookaside = module.params['lookaside']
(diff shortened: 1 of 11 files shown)


Odilhao (approved):
Approved, tested locally and it worked against the PRs that are broken atm on foreman-packaging

PR #445, opened 2026-08-16
Add --branch option to obal update --commit

zjhuntin:
## Summary

- Adds a `--branch BRANCH` CLI option to `obal update` that lets the caller specify the branch name used when `--commit` is set
- If the specified branch already exists it is checked out (rather than created), so multiple packages can be committed onto the same branch across successive `obal update` runs
- Falls back to the existing auto-generated branch name (`<current-branch>-update-<package>-<version>`) when `--branch` is not provided

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ogajduse:
What is the problem this PR is trying to solve?

Edit: Is it that _Obal cannot update more than one package in a single branch_?

obal/data/roles/git_branch_and_commit/tasks/main.yml
```diff
@@ -14,12 +14,35 @@
...
-  command: "git checkout -b {{ git_branch.stdout }}-update-{{ inventory_hostname }}-{{ updated_version.stdout }}"
+  command: "git checkout -b {{ target_branch }}"
+  args:
+    chdir: "{{ inventory_dir }}"
+  run_once: true
+  changed_when: true
+  when: branch_exists.rc != 0
+
+- name: 'Checkout existing git branch'
+  command: "git checkout {{ target_branch }}"
```
ogajduse, line 40: `git rev-parse --verify {{ target_branch }}` and the subsequent `git checkout {{ target_branch }}` aren't scoped to `refs/heads/`, so they'll also match a tag or other ref with the same name, resolving it and checking out a detached HEAD instead of failing or creating the branch.  The follow-up git commit -a then lands on no branch.

This is not a real-world risk today, but it is still an incorrect primitive for "does this branch exist".

```suggestion
- name: 'Check if branch exists'
  command: "git show-ref --verify --quiet refs/heads/{{ target_branch }}"
  args:
    chdir: "{{ inventory_dir }}"
  register: branch_exists
  failed_when: false
  changed_when: false
  run_once: true

- name: 'Create git branch'
  command: "git switch -c {{ target_branch }}"
  args:
    chdir: "{{ inventory_dir }}"
  run_once: true
  changed_when: true
  when: branch_exists.rc != 0

- name: 'Checkout existing git branch'
  command: "git switch {{ target_branch }}"
```

obal/data/roles/git_branch_and_commit/tasks/main.yml
```diff
@@ -14,12 +14,35 @@
   register: git_branch
   run_once: true
 
+- name: 'Set target branch name'
+  set_fact:
+    target_branch: "{{ branch | default(git_branch.stdout + '-update-' + inventory_hostname + '-' + updated_version.stdout) }}"
+  run_once: true
+
+- name: 'Check if branch exists'
```
ogajduse, line 22: The "commit multiple packages onto the same branch across successive runs" use case this PR is meant to enable breaks as soon as the caller resets the worktree to the base branch between runs — which is what both existing callers already do unconditionally before every obal update invocation (`update_packages.py's update_package()` and actions' `UpdatePackagesAction._update_package()` both run git checkout origin/<branch> first, every time). The update play mutates the spec/sources before this role decides which branch to land on, so if the target branch already has a differing committed version of that file, the checkout aborts.

```
# Run 1: obal update nodejs-lodash --version 4.18.2 --commit --branch batch-run
$ git checkout -b batch-run
$ git commit -am "Update nodejs-lodash to 4.18.2"
$ git log --oneline batch-run -1
9be1543 Update nodejs-lodash to 4.18.2

# Automation resets before the next call, same as brook/actions do today:
$ git checkout SATELLITE-STREAM

# Run 2: obal update nodejs-lodash --version 4.18.3 --commit --branch batch-run
# (update playbook bumps the spec in the worktree first, uncommitted)
$ git status --short
 M packages/foreman/nodejs-lodash/nodejs-lodash.spec

# obal's "Check if branch exists":
$ git rev-parse --verify batch-run
9be154311e95572a6c18bbc0b9ec382eead66ac3
rc=0

# obal's "Checkout existing git branch":
$ git checkout batch-run
error: Your local changes to the following files would be overwritten by checkout:
      packages/foreman/nodejs-lodash/nodejs-lodash.spec
Please commit your changes or stash them before you switch branches.
Aborting
```

zjhuntin:
The thing I'm trying to solve with this is updating many packages at once, but having a indiviudal commit for each package update, which I think works and often feels correct.  It often causes problems because every time I update multiple packages the branch name that is created gets in the way.

